PR #5, opened 2016-08-07
Submitting a proposal to discuss concerns of the Ensime open source project. 

bvenners:


SethTisue:
at the risk of being pedantic, this seems like either three or four separate agenda items (depending on whether the two presentation-compiler issues get lumped together)

bvenners:
My thought was to just go over them all briefly in one shot to bring them to the attention of the Scala Center and its advisory board and have some discussion. But I'm happy to refactor into multiple proposals if that's preferred.

propensive:
Hi @bvenners, @SethTisue,

@heathermiller and I decided not to have this as a proposal during the meeting tomorrow, but as a separate agenda item for discussion, from which new proposals may arise.

bvenners:
@propensive That sounds like a better approach.

PR #8, opened 2016-08-09
proposal/007-dotty-migration-path

timperrett:


proposals/007-dotty-migration-path.md
```diff
@@ -0,0 +1,28 @@
+
+# SCP-007: Migration path to Dotty for Scalac users.
```
propensive: According to the minutes, this should be SCP-002

PR #10, opened 2016-08-30
correct license wording in advisory board minutes

SethTisue:
Scala's license is absolutely standard BSD 3-clause, not merely
"BSD-style"

### DIFF
--- a/minutes/002-2016-q3.md
+++ b/minutes/002-2016-q3.md
@@ -226,7 +226,7 @@ contributors.
 
 The library reimplementation work is done "clean room" style, using
 Oracle documentation but without ever consulting Oracle code (since
-their license is GPL but Scala Native is BSD-style).  There was some
+their license is GPL but Scala Native is BSD 3-clause).  There was some
 discussion among the board about the legal issues around clean room
 implementations. Dag suggested looking at the
 [contributor questionnaire from the Apache Harmony project](http://harmony.apache.org/auth_cont_quest.html)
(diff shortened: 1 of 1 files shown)


SethTisue:
I will submit a separate PR in the scala-lang repo adding the "Scala is licensed under the BSD 3-Clause License " wording from https://github.com/scala/scala/blob/2.12.x/doc/LICENSE.md to the top of http://www.scala-lang.org/license.html

BennyHill:
Is it important that the scala license has not always been "pure" BSD, it most definitely used to be a modified verion - see  https://github.com/scala/scala/commit/bf93057e2eb41629f69657ea4a60acac0b824fce ?

My main reason for highlighting this is to show that changing a license is not actually has hard as suggested in the minutes, so not an obstacle for migratation to apache 2.0, as @fommil suggested by proxy, and @milessabin in https://github.com/typelevel/scala#relationship-with-lightbend-scala

SethTisue:
the previous change only standardized the wording without altering the meaning. it seems to me that success there has no bearing on what it would take to make a substantive change.

fommil:
changing a BSD 3 Clause is trivial. EPFL agreeing to use Apache 2.0 going forward is the problem: they'd have to clear releases with their patent office.

BennyHill:
ahh, OK. I was just curious, that's all. Thx :)

minutes/002-2016-q3.md
```diff
@@ -321,8 +321,8 @@ sbt).  These topics could eventually become concrete proposals.
 
 #### The Scala license
 
-Bill summarized Sam's licensing concern as follows.  Scala has
-[its own BSD-style license](http://www.scala-lang.org/license.html),
+Bill summarized Sam's licensing concern as follows.  Scala uses
+[the BSD 3-clause license](http://www.scala-lang.org/license.html),
```
fommil: "a 3-Clause Simplified BSD License" if you want to be exact :wink: http://ensime.github.io/contributing/center/#licensing

PR #12, opened 2016-11-20
[000] Improve user experience for builds that use only direct dependencies

stuhood:


stuhood:
This proposal should now be ready for review.

@xeno-by pointed out that https://github.com/scala/scala-dev/issues/275 by @retronym is related, so that issue is now linked.

proposals/000-ensure-compilation-scalability.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Abstract
+
+The most commonly discussed angle of attack for improving the Scala build experience is improving the speed of the compiler. Instead, this document addresses a different angle: ensuring that it is possible to scale multiple, parallel (even remote) invocations of the compiler.
+
+When building very large Scala projects, the presence/necessity of transitive build dependencies on the compile classpath represents a very real limit on the ability to distribute compilation: shipping 2000 transitive dependencies to a cluster of remote compiler instances is significantly more challenging than shipping 20 direct dependencies. Additionally, over-declaring dependencies causes cache keys for build artifacts to be invalidated more often than is strictly necessary.
+
+scalac generally does not require the entire transitive classpath of a library to compile it, but its requirements are not formally defined, and in many cases are non-intuitive. This uncertainty makes it difficult to accurately declare the direct/"strict" dependencies of a library... a generally accepted build-hygiene best practice with great benefits for scalability (and straight-line performance).
```
xeno-by: I think it would be useful to explicitly define what a strict dependency is.

retronym:
The name of this proposal suggests a different goal to me (e.g. make scalac use multiple threads, or reduce contention between multiple instances of scalac in a single JVM).

How about: "specify and test compiler's robustness against absent classpath entries"?

proposals/000-ensure-compilation-scalability.md
```diff
@@ -0,0 +1,45 @@
...
+
+One example case (of many) where an unimported dependency is required is when a compilation unit `org.c` uses a class `A` defined in a separate compilation unit `org.a`, which implements a trait `B` defined in a compilation unit `org.b`. While compiling `org.c` with `org.a` on the classpath (but not `org.b`), scalac emits an error like the following:
+
+    [error] missing or invalid dependency detected while loading class file 'A.class'.
+    [error] Could not access term b in package org,
+    [error] because it (or its dependencies) are missing. Check your build definition for
+    [error] missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
+    [error] A full rebuild may help if 'A.class' was compiled against an incompatible version of org.
```
xeno-by: How about additing a dedicated section for the example. Something like:

## Example

One example case (of many) where a transitive dependency is required even when not used directly involves inheriting from a trait.

```
// A.scala
trait A

// B.scala
class B extends A

// C.scala
object C extends App {
  println(new B)
}
```

In this arrangement, `C` depends on `B` which depends on `A`. Even though `C` doesn't use `A` directly, `A` still has to be on the classpath, otherwise a compilation error occurs.

```
$ scalac -d a A.scala

$ scalac -d b -cp a B.scala

$ scalac -d c -cp b C.scala
error: missing or invalid dependency detected while loading class file 'B.class'.
Could not access type A in package <empty>,
because it (or its dependencies) are missing. Check your build definition for
missing or conflicting dependencies. (Re-run with `-Ylog-classpath` to see the problematic classpath.)
A full rebuild may help if 'B.class' was compiled against an incompatible version of <empty>.
one error found
```

Another recent example of this type of mismatch is found in [scala/scala-dev#275](https://github.com/scala/scala-dev/issues/275).

retronym: `javac` requires all base classes to be present on the compilation classpath, its unlikely that we'll outdo them in this regard.

```
tail -n 10 sandbox/{A,Test}.java && javac -d . sandbox/A.java && rm A.class; javac -d . sandbox/Test.java
==> sandbox/A.java <==
class A {}
class B extends A {}

==> sandbox/Test.java <==
class Test {
    public static void main(String[] args) {
        System.out.println(new B());
    }
}
sandbox/Test.java:3: error: cannot access A
        System.out.println(new B());
                  ^
  class file for A not found
1 error
```

The standard use case for stub symbols that I'm aware of are:

```
⚡ tail -n 10 sandbox/{A,Test}.java && javac -d . sandbox/A.java && rm A.class; javac -d . sandbox/Test.java
==> sandbox/A.java <==
class A {}
class B {
    A unused() { return new A();}
    void used() {}
}

==> sandbox/Test.java <==
class Test {
    public static void main(String[] args) {
        new B().used();
    }
}
```

Or:

```
⚡ tail -n 10 sandbox/{A,Test}.java && javac -d . sandbox/A.java && rm A.class; javac -d . sandbox/Test.java
==> sandbox/A.java <==
@interface A {}
class B {
    @A void used() {}
}

==> sandbox/Test.java <==
class Test {
    public static void main(String[] args) {
        new B().used();
    }
}
```

retronym: It is important to test variations of these that include:

  - entire packages are missing (rather than just some class missing from a package with some other entries)
  - nested packages (e.g. `p1.p2.p3.Missing` where `p2` and `p3` themselves are missing packages)
  - References to the missing classes from Scala and Java defined classfiles (we typically need to implement stub symbol handling in both places in the compiler)

stuhood: @retronym : Thanks for the examples!

Note that the javac requirement you mention is (possibly?) a relatively new change: JDK8 changed things in this regard: see the snippet in https://github.com/sbt/zinc/issues/148 about this.

retronym:
@stuhood Interesting! My example involves a superclass, not an interface, which isn't mentioned by that release note.

proposals/000-ensure-compilation-scalability.md
```diff
@@ -1,26 +1,38 @@
...
 
 ## Proposer
 
-Proposed by Stu Hood, Twitter, November 2016
+Proposed by Stu Hood, Eugene Burmako - Twitter - November 2016
 
 ## Abstract
 
-The most commonly discussed angle of attack for improving the Scala build experience is improving the speed of the compiler. Instead, this document addresses a different angle: ensuring that it is possible to scale multiple, parallel (even remote) invocations of the compiler.
+The most commonly discussed angle of attack for improving the Scala build experience is improving the speed of the compiler. Instead, this document addresses a different angle: improving the experience of using only "direct" dependencies in builds, in order to improve cache hit rates and unblock scaling multiple remote invocations of the compiler.
```
xeno-by: this document => this proposal?

xeno-by: If we're gonna talk about other techniques enabled by minimization, we should mention them here, too.

xeno-by: Also, along with speeding up the compiler, we should probably mention incrementality as well.

PR #15, opened 2016-12-01
Replace SLIP by SIP

jvican:
SLIP doesn't exist anymore. As the SIP Committee will choose what goes into the core and what doesn't, it makes sense that if a proposal is submitted is a SIP and not a SLIP (nor a SPP).

### DIFF
--- a/proposals/011-collections.md
+++ b/proposals/011-collections.md
@@ -16,7 +16,7 @@ The Scala collections library underwent a major redesign in Scala 2.8. While the
 
 Lightbend will contribute a significant amount of engineering time to lead and help implement the project with the goal of integrating the library into Scala 2.13. Due to the size of the effort and the large number of interested parties, the Scala Center should also be involved in the design and implementation of the new collections library. We expect the following activities as part of the project:
 
-- Define the new design, following the appropriate Scala library improvement process (SLIP):
+- Define the new design, following the appropriate Scala Improvement process (SIP):
   - Solicit and incorporate feedback from the existing strawman proposals, the community and commercial users.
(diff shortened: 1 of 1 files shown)


heathermiller:
I think it's up to discussion who gets to decide what goes into core and who doesn't. This might be better-phrased as a suggestion that the SIP committee is the one to decide what's a part of Scala Core and what isn't than a PR.

jvican:
The SIP Process specifies that the Scala Core is a reponsability of the SIP Committee:

> Responsibilities of the members
> 1. Review the proposals they are assigned to. The Process Lead will always notify them two weeks in advance, at minimum.
> 2. Play a role in the discussions, learn in advance about the topic if needed, and make up their mind in the voting process.
> 3. **Decide which utilities should be inside the core module and are required by the compiler**. The goal is to shrink it over time, and, where possible, move modules to the platform, that will be managed by the SLIP process.

heathermiller:
Yes, this is a document that you wrote where you've assigned these responsibilities to the SIP committee. However, this is an important point that we make sure everyone agrees with.

heathermiller:
I believe most everyone else was not aware that this was specified in your SIP document. Most notably, this wasn't clear to the Scala compiler team at Lightbend (as evidenced by the advisory board meeting yesterday).

SethTisue:
This is exactly one of the things people were (implicitly) asking about when they asked if SPP replaces SLIP.

SethTisue:
(I actually think it makes sense to have the core library under SIP, but I haven't had much time to think about it yet.)

PR #20, opened 2017-02-22
Proposal: Prioritize and collate sbt/scala/dbuild/community-build issues for community projects

larsrh:
Review by @BennyHill – would you mind providing a brief abstract as well?

(NB, this is mostly copy provided by @BennyHill)

### DIFF
--- a/proposals/010-community-build.md
+++ b/proposals/010-community-build.md
@@ -1,0 +1,30 @@
+# Prioritize and collate sbt/scala/dbuild/community-build issues for community projects
+
+## Proposer
+
+Proposed by @BennyHill (and @larsrh), Typelevel, 2017-02-22
+
+## Abstract
+
+TBD
+
+## Proposal
+
+We would like to propose that the Scala Center maintain a community driven list of the open sbt/scala/dbuild/community-build issues that they would like priority given too. This could be as simple as a GitHub-driven list, but we would like to also suggest that the real value the SC could add is identifying issues that impact many projects and the general Scala user experience, even if the individual issues themselves are not heavily voted for.
+
+Indeed, this could also progress to having "Meta Issues", groups of issues from many projects that on their own do not seem important for the individual projects, but are when viewed together, across projects.
+
+Note that this does not stipulate that the SC actually do the work for the issues, but rather help focus issues that might otherwise get lost. By highlighting such, perhaps the community are motivated to do the work, perhaps some SC board members have Lightbend (LB) subscriptions and lobby other LB subscribers to vote for them.
+
+## Cost
+
+We expect this work to consist mostly of triaging and keeping track of issues; also responding to community feedback. We estimate 0–0.25 FTE. 
+
+## Timescales
+
+Continuous.
+
+## References
+
+* [typelevel/general#62](https://github.com/typelevel/general/issues/62)
+* [typelevel/general#45](https://github.com/typelevel/general/issues/45)
--- a/proposals/010-community-build.md
+++ b/proposals/010-community-build.md
@@ -16,6 +16,31 @@ Indeed, this could also progress to having "Meta Issues", groups of issues from 
 
 Note that this does not stipulate that the SC actually do the work for the issues, but rather help focus issues that might otherwise get lost. By highlighting such, perhaps the community are motivated to do the work, perhaps some SC board members have Lightbend (LB) subscriptions and lobby other LB subscribers to vote for them.
 
+Examples of issues that are are spread across projects include:
+
+- Defining custom community builds (CB)
+  - Sharing configurations from other CB's
+  - Getting shared configs via scaladex
+  - Running then locally (via catalysts-docker)
+  - Promoting these tools - if a project does not know a solution exists or is fit for a certain task, they will not use it. 
+- Pattern matching vals in sbt
+  - An old issue, advocated for fixing by many - see  https://github.com/sbt/sbt/issues/2290
+https://github.com/sbt/sbt/issues/1661 - but perhaps not a high priortiy for business aligned projects
+  - relates to "-Y rangepos" option in scalac. So the SBT fix also requires a scalac fix. 
+  - The general issue here has also been a recurring issue in scoverage, probably the main test coverage tool for scala.
+- Cross building projects
+  - This is not a standard SBT feature
+  - Scala.js added cross project support for JVM/JS
+  - scala-native has now taken this over, extending the Scala.js implementation
+  - Typelevel Scala added its own platform type
+  - One method of implementing this is using  "Pattern matching vals in sbt"
+  - Adding support of multiple version of libraries makes the current API almost unusable - who should be responsible for this? (How code is maintained for different library versions is a separate concern)
+- Rewriting scala code to support multiple library versions
+  - scala-meta and scalafix can be leveraged, and are in [catz-cradle](https://github.com/BennyHill/catz-cradle)
+  - custom community builds could test rewrites on existing projects, perhaps via scaladex
+  - Difficult to implement and test as SBT does not have full support for cross building....
+
+
 ## Cost
 
 We expect this work to consist mostly of triaging and keeping track of issues; also responding to community feedback. We estimate 0–0.25 FTE. 


SethTisue:
> this could also progress to having "Meta Issues", groups of issues from many projects that on their own do not seem important for the individual projects, but are when viewed together, across projects

some of these might logically progress to become Scala Center proposals

SethTisue:
> community driven list

I think my main question about this proposal is what's envisioned here in terms of process.

heathermiller:
Hey guys, first off, thanks a lot for pooling effort to come up with a proposal for the AB. I want to get this into a shape where we can identify concrete work items so we can ultimately take on some of these issues. 

Only problem is that I'm a bit confused by the proposal. In particular, the list of issues spread across community builds reads to me like several very different potential proposals lumped together. The main proposals I see are:

1. Defining custom community builds (CB)
2. Pattern matching vals in sbt
3. Cross building projects
4. Rewriting scala code to support multiple library versions

These could logically each be their own proposals. 

Since the AB meetings are heavily streamlined, it would be good to flesh out this proposal a bit more ASAP so I can figure out staffing (who could potentially do this work), and how much time it would take, so we could vote on it in the next AB meeting next week.

For what it's worth, @MasseGuillaume has already done work on (3), as part of one of the Scala Native proposals, so, cross-building for Scala Native projects. Just thinking out loud here, perhaps that 

> Indeed, this could also progress to having "Meta Issues", groups of issues from many projects that on their own do not seem important for the individual projects, but are when viewed together, across projects.

Very good point! I guess this would fall into the "defining custom community builds" bullet point? I'm still a bit confused/unclear on what we could specifically do to achieve this. 

I like the ideas of making it easier to somehow share this information, maybe via Scaladex as you suggested, but I still want to drill down into what exactly meta issues are, and how we could potentially socially connect all interested Scala library maintainers in order to arrive at an aggregation of these. 

On point (4), @olafurpg has already been working towards this goal as part of the Scalafix project, with a bunch of help and feedback from @ShaneDelmore. Given that, couldn't this sort of rewriting work be lumped into the existing AB proposal that has spawned Scalafix?

Just trying to bring more clarity to what concretely we can do, so we can streamline it and present it during the AB meeting, and I can give accurate judgements of who can do what and how long it should take.

proposals/010-community-build.md
```diff
@@ -0,0 +1,55 @@
...
+
+We would like to propose that the Scala Center maintain a community driven list of the open sbt/scala/dbuild/community-build issues that they would like priority given too. This could be as simple as a GitHub-driven list, but we would like to also suggest that the real value the SC could add is identifying issues that impact many projects and the general Scala user experience, even if the individual issues themselves are not heavily voted for.
+
+Indeed, this could also progress to having "Meta Issues", groups of issues from many projects that on their own do not seem important for the individual projects, but are when viewed together, across projects.
+
+Note that this does not stipulate that the SC actually do the work for the issues, but rather help focus issues that might otherwise get lost. By highlighting such, perhaps the community are motivated to do the work, perhaps some SC board members have Lightbend (LB) subscriptions and lobby other LB subscribers to vote for them.
+
+Examples of issues that are are spread across projects include:
+
+- Defining custom community builds (CB)
```
olafurpg: I think this proposal could benefit from clarifying with what is meant by a "Community build".

In my view, community builds come in all sorts of shapes and sizes with different needs. If we boil it down, they share the following traits:

1. a corpus of non-synthetic projects that depend on your project, and 
2. some system to automatically run/test this corpus against the latest snapshot of your project

Challenges of maintaining a community build include making

- the corpus representative for the inputs you want to test.
- the system as straightforward as possible (fast + simple setup) to run

In the case of scala/scala, the corpus is the manually curated list of repos+test-targets in scala/community-builds and the system is dbuild. dbuils solves an extra tricky problem since the corpus creates a deep tree of dependencies and dbuild needs to build the tree from bottom-up.

If you are so fortunate that the corpus (part 1) is not a complex web of dependencies, you can get away with much less advanced system than dbuild. For example in every scalafmt PR, the CI downloads a 20mb zip file of source files, unzips under `target/` and reformat all ~3 million lines of Scala code. It's simple, fast and very effective at finding bugs.

MasseGuillaume:
@heathermiller 
Regarding 
> 2. Pattern matching vals in sbt
> 3. Cross building projects

This is how the api look like for [sbt-crossproject](https://github.com/scala-native/sbt-crossproject/blob/master/sbt-cross-test/src/sbt-test/new-api/aggregates/build.sbt#L3).

Our solution is source compatible with the Scala.js solution. Out goal was not to improve the api but to be compatible with what Scala.js already have.

In typelevel/cats they end up with an explosion of projects: https://github.com/typelevel/cats/blob/master/build.sbt#L182

dwijnand:
@MasseGuillaume Ideally

```scala
lazy val bar = crossProject(JSPlatform, JVMPlatform, NativePlatform)
  .crossType(CrossType.Pure)
  .settings(scalaVersion := "2.11.8")
  .nativeSettings(resolvers += Resolver.sonatypeRepo("snapshots"))

lazy val barJS     = bar.js
lazy val barJVM    = bar.jvm
lazy val barNative = bar.native
```

could be something like

```scala
lazy val (bar, barJS, barJVM, barNative) = crossProject(JSPlatform, JVMPlatform, NativePlatform)
  .crossType(CrossType.Pure)
  .settings(scalaVersion := "2.11.8")
  .nativeSettings(resolvers += Resolver.sonatypeRepo("snapshots"))
  .crossAndSplitAll
```

MasseGuillaume:
@dwijnand that would be cool. 

I think the difficulty is that we have a build matrix with multiple dimensions
* targets: Dotty, Scala.js, Scala-Native, Typelevel's Scala, ...
* scala versions: 2.10, 2.11, 2.12, ...
* architectures: x64, x86, ...
* dependencies: spark 2.x, spark 1.x, ...

We can basically create as many project permutation than we have inputs. This is one problem highlighted above.

Another difficulty is to do projections. For example, test all cats' Jvm projects.

In Scastie, I assemble those inputs manually: https://github.com/scalacenter/scastie/blob/master/build.sbt#L62

proposals/010-community-build.md
BennyHill: I would suggest that if "the corpus (part 1) is not a complex web of dependencies" then you don't really need a community build, but that you _might_ benefit from the tools in anyway.

BennyHill:
All, So the examples given are just specific examples that I have been involved with, and working with, too. The main point of the proposal is that SC promote and establish a place( eg a list) to pull these things together, especially where there are overlapping concerns.

Stand back and look at the comments above, that are actually related to the examples, not the proposal: We have already brought together a discussion from different stakeholders and different projects! Is this proof/evidence that this proposal works?

BennyHill:
Perhaps it would help to provide and discuss an abstract. Version 1 -

> Scala projects already have good tools to help their communities communicate, tools such as github, gitter, mailing lists - it's a long list. This proposal is for the Scala Center to establish a focus point to promote discussion and collaboration _between_ projects.

proposals/010-community-build.md
olafurpg: If so, I would make that assumption explicit and maybe also include an estimate of how many projects in the community require a CB with a complex web of dependencies.

> but that you might benefit from the tools in anyway.

Do you have any specific tools in mind? One example might help clarify this part.

BennyHill:
@heathermiller In reference to your main points, in case you were not aware

1. CB - https://github.com/scala/community-builds/pull/478 and [catalysts-docker](https://github.com/typelevel/catalysts-docker)
2. Pattern matching vals in sbt - https://gitter.im/sbt/sbt?at=58a71dedaa800ee52cab57a0 onwards
3. Cross building:  https://github.com/typelevel/sbt-tls-crossproject
4. Rewriting scala code to support multiple library versions - https://github.com/typelevel/catz-cradle

So this is how/why I know how these projects overlap :smile: 

And also from https://github.com/typelevel/sbt-catalysts, including [the typelevel dependencies](https://github.com/typelevel/sbt-catalysts/blob/master/src/main/scala/org/typelevel/TypelevelDeps.scala)

proposals/010-community-build.md
BennyHill: That is a good point, but the point of the proposal is too promote an "area" where these topics are discussed, this is just one example.

[ But to answer your question - scala/community-builds#478 and [catalysts-docker](https://github.com/typelevel/catalysts-docker) ]

BennyHill: >  I think it could benefit from clarifying what is meant by a "Community build".

Again. I agree that this is needed...but not here, rather in the "new place" that this proposal proposes. These are good comments that you have made - and support, by implication, the proposal :smile:

olafurpg:
> This proposal is for the Scala Center to establish a focus point to promote discussion and collaboration between projects

I think this is a great idea! I would try to elaborate on it more in the abstract/proposal and include examples of concrete measures that would move the needle towards that goal.

> Re: pattern matching vals in sbt

(I understand this proposal is not about concrete issues, but I still want to pitch my 2 cents on this topic)
I think this is a suboptimal solution to the wrong problem. Firstly, I think the solution is suboptimal because there are few guards to prevent users from mixing the order in the pattern match. Secondly, it's the wrong problem to solve since I think sbt's `project` concept should rather be extended to handle multiple output target in the same way as bazel/pants targets do. For example, a single protobuf target in bazel can be consumed by java/c++/python/... targets. The `coreJVM/coreJS` solution is a workaround for sbt's limitation.

BennyHill:
@olafurpg So to continue your issue, and what this proposal would like to address - 

> sbt's project concept should rather be extended

From the chat I referenced ^^ 

> "so likely the fix would need to come from the community, unless Lightbend subscribers can lobby for it"

How do we get  Lightbend subscribers in on the discussion?

proposals/010-community-build.md
olafurpg: I think CBs are a very effective form of testing. I totally agree that it would be great if we could lower the barrier-of-entry for project maintainers to setup their own CBs 😄

BennyHill:
@larsrh There's an abstract there ^^, somewhere :smile:

heathermiller:
> The main point of the proposal is that SC promote and establish a place( eg a list) to pull these things together, especially where there are overlapping concerns.

While it might be clear to you that this is what the proposal is about, from the text as-is, it wasn't clear to us, especially given the list of so many different projects in the proposal. For us to present this at the AB meeting, this proposal really needs to be streamlined and clarified. A good example to go off of is the [collections proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/007-collections.md). As-is, it's a little too scattered, and it's missing pieces; e.g, an abstract. Perhaps @larsrh will be able to better translate/clarify in the actual AB meeting, but I would seriously encourage someone @larsrh/@BennyHill to do a simplification/streamlining/clarification pass on this ahead of the meeting.

BennyHill:
@heathermiller In my [original issue to TL](https://github.com/typelevel/general/issues/62), I actually separated out the proposal and the examples as I was concerned the two would get mixed up - looks like we should do the same here. Remove the examples, @larsrh can discuss them at the meeting.

Would this be a good start?

heathermiller:
@BennyHill Yep, that original text at the top of the TL issue is a good start! It'd be good to just focus on that in the rest of the proposal, and to flesh out concrete actionable tasks that you guys think the SC can do in line with this original proposal. If you have a look at the [collections proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/007-collections.md), that's exactly what it does. It starts with a motivation, and then it moves on to concrete goals as well as an idea of a concrete _thing_ (in this case a library) that the SC would produce. Try to do the same thing here.

My trouble with this proposal as it currently is is that there are so many ideas, so many conversations floating around, links to Gitter chats, different issue trackers, etc, that it's really hard for me to see what the concrete work is that the SC can do. If you can make that clearer ASAP, this would probably get accepted at the AB meeting next week. 🙂 

So, if I were you, I'd focus on getting this proposal into the expected shape (again, have a look at the [collections proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/007-collections.md) as an example). Focus on expanding the proposal text in the [top of your TL issue](https://github.com/typelevel/general/issues/62) with some concrete tasks that the SC can do, so we can discuss it and make a decision whether or not the SC should undertake this at the AB meeting.

BennyHill:
@heathermiller OK, all fair points - and many thanks for taking the time to push for these changes _before_ the meeting!

BennyHill:
In fact..... "there are so many ideas, so many conversations floating around, links to Gitter chats, different issue trackers, etc, that it's really hard for" _is_  what this proposal would like to address

heathermiller:
@BennyHill No worries! I want to see this _in_ the upcoming meeting :) If you'd like, feel free to work in this PR. We're happy to provide feedback or to inundate you with questions where we're confused 😛

dwijnand:
@olafurpg 

> I think this is a suboptimal solution to the wrong problem. Firstly, I think the solution is suboptimal because there are few guards to prevent users from mixing the order in the pattern match. Secondly, it's the wrong problem to solve since I think sbt's project concept should rather be extended to handle multiple output target in the same way as bazel/pants targets do. For example, a single protobuf target in bazel can be consumed by java/c++/python/... targets. The coreJVM/coreJS solution is a workaround for sbt limitations.

Sure. But I also just want to remove the restriction in sbt files and make it more similar to actual Scala. (oh btw it's a distinction scalameta dialects doesn't know about)

larsrh:
Apologies for not having the whole thread closely, but as far as I understand, the examples should be removed and I should mention them in the meeting? Anything else I have missed?

heathermiller:
@larsrh I guess the main todo is that this proposal needs to be worked out a bit more. So if the examples are removed, sure, but we need to figure out more concretely what the Scala Center can do to help which problem–there seem to be several different problem areas outlined, and myself and others are still a bit confused by this proposal.

As far as next steps, this comment is probably clearest about next steps: 
https://github.com/scalacenter/advisoryboard/pull/20#issuecomment-281966845

I'd really like to streamline this so we discuss it and take some action at next week's AB meeting!

larsrh:
Right, thanks. @BennyHill, could you perhaps send me some updates along those lines? I'm still on the conference circuit until Sunday so unlikely to find time to fix the proposal.

propensive:
We won't be able to accept this as a proposal as-is for tomorrow's meeting, but @larsrh will get an opportunity to raise it in the meeting, and there will be an opportunity to discuss it there. I hope that some concrete proposals can arise from that discussion, which can become proposals at the next meeting. I think deferring is the best course of action, as voting (for *or* against) an unready proposal isn't good for anyone.

larsrh:
@propensive Agreed.

Sorry everyone that I didn't manage to improve the proposal in time.

propensive:
@larsrh @BennyHill Is there any interest in having this proposal discussed at the next meeting, on 31 May?

BennyHill:
@propensive Thanks for the prompt, unfortunately I currently have no free time for this.

@larsrh Perhaps you have some time to massage ^^ into a proposal?

larsrh:
Unfortunately not, sorry.

propensive:
Ok, there's no problem with the PR remaining open (for now) until it's ready.

propensive:
Quarterly ping, @larsrh. ;)

larsrh:
@propensive What's the deadline for the next meeting?

BennyHill:
btw: see https://contributors.scala-lang.org/t/continuing-or-dropping-scala-2-10-maintenance-in-the-ecosystem/1013/16 as a more recent, related example

BennyHill:
and another one, https://gitter.im/scala/community-builds?at=58e495fd8bb56c2d11783acd 

(my highlighting)

> Ideally we could get the community build to tell us what warnings a PR changes. We have essentially __no tools for analyzing the build logs__.

propensive:
@larsrh @BennyHill This is your quarterly ping! ;) Deadline for submissions for the next meeting is midnight (UTC) on 27 November.

propensive:
By the way, I'm tempted, if we don't get this proposal into the next meeting, to close the PR. That's not to say in couldn't be reopened again, but it's failed to make (I think) three meetings now, so I'd rather it didn't hang around in this state forever.

larsrh:
Neither of us has time to work on this now. I think it can be closed.

BennyHill:
This PR is long closed, but I think it is worth adding https://contributors.scala-lang.org/t/alternative-scalajs-scala-native-distribution-mechanisms that came up in September 2017.

Not identical, but alot in common - just sayin. And FTR, the  hardest issues in cats and related are generally scala `Mx`releases, not cross platform builds. The reason I add this point is not to throw stones, but to point out that everyone looses out until this is streamlined

PR #23, opened 2017-04-17
Improving SBT learning material

propensive:


SethTisue:
speaking only for myself, I would like to see the main Scala websites in good shape — in other words, for SCP-008 (https://github.com/scalacenter/advisoryboard/blob/master/proposals/008-websites.md) to be more or less complete — before an sbt doc/teaching effort was embarked on.

Some SCP-008 work is happening, but a lot hasn't even been started, and even the work items that are happening (e.g. the tour of Scala, and the scala-lang changes from 47 Degrees) aren't actually across the finish line yet.

propensive:
Maybe this could be incorporated into the timescales? I said shortly after SBT 1.0 is released, but I'm not sure what the timeframe on that is, other than "soon".

heathermiller:
I don't intend to be a stickler here, but I'm not sure that it's correct for the chairman of the advisory board to be submitting proposals. To be correct and not to give preferential treatment to anyone, I think we should do this as specified in the AB documents – this is a proposal that should be made by one of the two community representatives on the AB.

heathermiller:
As far as I recall, the rule clearly stipulate that only voting members of the AB can submit proposals. Therefore, Lars or Bill should be submitting this, not the Jon, I think.

proposals/xxx-improve-sbt-learning-materials.md
```diff
@@ -0,0 +1,63 @@
+# Improve Resources for Learning SBT
+
+## Proposer
+
+Proposed by Jon Pretty, Chairperson.
+
+FIXME: Need to decide whether I can make a proposal directly, or should go
+through a community representative.
```
heathermiller: Only voting members can make proposals, as far as I recall.

propensive: Yes, I think that's the correct process. I was intending to have the proposal submitted by someone else (which, I guess, could be any of the sponsors) before it gets merged as an actual proposal.

larsrh:
Happy to give a :+1: to this and act as the proposer.

heathermiller:
Sounds good @larsrh! 

So, just to go according to protocol then, I guess @larsrh should be mentioned as the submitter of the proposal in line 5 (maybe along with the text, "original proposer: Jonathan Pretty"). @propensive could you make this tweak? Otherwise LGTM!

larsrh:
> original proposer: Jonathan Pretty

Don't forget the middle name 😆

propensive:
Thank you @larsrh and @heathermiller!

heathermiller:
Merci Jon. LGTM!

SethTisue:
Was there a Twitter thread on this? I feel like there was, but I couldn't find it.

PR #25, opened 2017-04-27
clarify a position

fommil:
It is not completely obvious from the sentence to whom the advice was given. Of course, EPFL lawyers / commercialisation departments can only advise their clients, and are not at liberty to advise the Scala community. This small amendment clarifies that the advise was given from the EPFL perspective.

### DIFF
--- a/minutes/002-2016-q3.md
+++ b/minutes/002-2016-q3.md
@@ -334,7 +334,7 @@ be changed to a Apache 2 license?
 
 Heather said she forwarded Sam's questions to the EPFL technology
 transfer office and in their opinion, having the separate CLA doesn't
-cause any major legal loopholes or problems.
+cause any major legal loopholes or problems for EPFL.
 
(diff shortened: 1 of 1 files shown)


SethTisue:
fwiw, the original text is a more accurate summary of _what was actually said at the meeting_ 

what the tech transfer office actually said to Heather, I of course don't know. so, up to Heather whether to merge.

S11001001:
The original summary is probably more accurate.  However, if the advice was legal advice (of the non-IANAL variety), then that only EPFL's interests are taken into consideration is implied, but readers may not know that.

If the advice was not legal advice, then there is probably no reason to change anything.

propensive:
To avoid this staying open forever, and for the reasons @SethTisue and @S11001001 mention, I'm going to opt not to merge this. @heathermiller can re-open and merge, if necessary!

PR #27, opened 2017-08-02
Added latest minutes, and updated recommendations, and attendees

propensive:


SethTisue:
propagated to scala.epfl.ch, where I was surprised to discover that a different version of the minutes had already been published: https://github.com/scala/scala.epfl.ch/commit/cc72521668dd0b2f10fed368b64b6ce7f0d5ad62

I don't think it's good to publish the notes in two different locations like this. there ought to be a single master copy.

SethTisue:
tweeted https://twitter.com/scala_lang/status/893520697947025408

PR #28, opened 2017-08-03
Proposal: sbt migration

larsrh:


fommil:
Note that there is a migration list of plugins https://github.com/sbt/sbt/wiki/sbt-1.x-plugin-migration

The community effort has been much better than I expected. Once the play plugin is upgraded, I suspect we'll see more people move companies move to sbt 1.0. I don't know who's going to upgrade the ensime plugin (it's a huge piece of work) but presumably somebody will step up if it's important to them. This work would have been very valuable in the last six months but by the time SC resource can be allocated, I suspect most of the work will have been done already by volunteers. Perhaps it is something to consider in advance when sbt 2.0 is on the horizon...

larsrh:
True, but there's also a long tail of builds to be upgraded to 1.0.x.

dwijnand:
I'm equally pleased by the uptake by the plugin community in supporting sbt 1.

In regards to upgrading project builds to sbt 1, while I (obviously) encourage adoption of sbt 1 at large, I personally am not sure if this is a good effort to direct the SC's resources towards.

larsrh:
> I personally am not sure if this is a good effort to direct the SC's resources towards.

Good point, maybe not.

fommil:
FYI ENSIME's sponsored developer is going to work on this before adding language server support and scalajs rewrite.

PR #30, opened 2017-08-21
Proposal: production ready scalamacros

andyscott:
This proposal is the collaborative effort of many, including members of 47 Degrees, SAP, and Twitter.

I am opening the proposal PR so that we can begin public discussion and make adjustments more easily. For the sake of tracking, I will address feedback with incremental commits which we can squash later, if desired.

/cc @raulraja @xeno-by @olafurpg


fommil:
getting scala.meta working in the REPL, presentation compiler (i.e. ensime and scalaide) and docs is essential. I basically can't even use the gsoc project I'm mentoring because it just crashes.

Beyond this, having a larger support for for comprehensions is essential for working with async / monadic code. I've had problems with it parsing it incorrectly.

In many ways, I wish scala.meta's parser *was* the parser. It feels like there is a lot of duplication going on with tree generation.

andyscott:
The note about support for macros on for-comprehensions is specifically for annotation macros. As far as I know, def macros should be able to handle trees for for-comprehensions.

proposals/014-production-ready-scalamacros.md
```diff
@@ -0,0 +1,38 @@
...
+
+No one is currently employed to work on scalamacros/scalamacros. This means that the project relies on volunteer effort to reach a production-ready status.
+
+## Proposal
+
+We propose that the Scala Center invests engineering resources to make scalamacros/scalamacros production ready so the community can migrate to this new macro system. This will ultimately enable better IDE support and ease migration to dotty.
+
+Focus should be put on the following tasks:
+
+- Finalize support for macro annotations with coverage for class/trait/object/def definitions and modifiers. This does not include support for more complicated syntactic structures such as patterns or for-comprehensions.
```
olafurpg: > This does not include support for **pattern-matching against** more complicated syntactic structures such as patterns or for-comprehensions.

Scalamacros won't have a problems with source file that contain for comprehensions (unlike scalameta/paradise). Macro authors will only not be able to pattern match on such structures, like 
```scala
@ val q"for (..$enumerations) yield $expr" = "for { x <- List(1 /* comment */) } yield x".parse[Stat].get
enumerations: List[Enumerator] = List(Enumerator.Generator(Pat.Var(Term.Name("x")), Term.Apply(Term.Name("List"), List(Lit.Int(1)))))
expr: Term = Term.Name("x")
```
```

File: proposals/014-production-ready-scalamacros.md [hunk 2]
```diff
@@ -0,0 +1,38 @@
...
+
+Unfortunately, Scala macros have also gained notoriety as an arcane and brittle technology. The most common criticisms of Scala macros concern their difficult metaprogramming API based on compiler internals as well as a lack of featureful editor support. Even five years after their introduction, traditional macros don’t expand properly in Intellij. This leads to a proliferation of spurious “red squiggle” errors, even in simple projects. As a result of these problems the language committee has decided to retire the current macro system in Scala 3.x.
+
+During the last couple years a new macro system has been developed to support both Scala 2.x and Scala 3.x. The new system is based on the platform-independent metaprogramming API Scalameta, which is designed to be easy to use and easy to support across multiple implementations of the Scala language. The result of this multi-year effort is hosted at https://github.com/scalamacros/scalamacros and its design is explained in detail in the paper  “Two approaches to portable macros”.
```
olafurpg: Missing link at the end: https://www.dropbox.com/s/2xzcczr3q77veg1/gestalt.pdf

proposals/014-production-ready-scalamacros.md
```diff
@@ -0,0 +1,38 @@
...
+
+This repository is currently a working prototype of the new macro system that features:
+Scalameta based syntactic and semantic APIs that cross-compile against Scala 2.12, 2.13 and Dotty. The corresponding library is quite slim, being less than 500Kb in size.
```
olafurpg: Add bullets?

proposals/014-production-ready-scalamacros.md
```diff
@@ -0,0 +1,38 @@
...
+Focus should be put on the following tasks:
+
+- Finalize support for macro annotations with coverage for class/trait/object/def definitions and modifiers. This does not include support for more complicated syntactic structures such as patterns or for-comprehensions.
+- Finish support for def macro materializers with full coverage for types. This includes support for the functionality provided by legacy whitebox macros-- specifically allowing def macros to inspect types at the macro callsite and return values of types not known before macro expansion.
+- Provide documentation for using the new macro system. This includes a migration guide from legacy macros, potentially with scalafix rewrites.
+
+
+## Cost
+
+We expect each of these items to require a minimum one month effort. Combined, we expect 4 to 5 months of a full-time hired engineer.
```
fommil: "cost" is supposed to be "external costs" not resource. @propensive needs one of these: https://youtu.be/BfA_Z7DpiWU?t=60

You could move the effort into timescale.

proposals/014-production-ready-scalamacros.md
dwijnand: There are two mentions of "Scala 3.x" and two of "Dotty". Assuming they mean the same thing, IMO the document should either (a) consistently say the same thing, or (b) alias one to the other before switching (such as "Scala 3.x (also known as Dotty)").

proposals/014-production-ready-scalamacros.md
fommil: I'd go so far as to say that it has to be @olafurpg who delivers on this. It's not really feasible to expect anybody except the core `scala.meta` developers to be able to do this.

heathermiller: SC resourcing and management is not the job of the proposers.

fommil: We've already had a failed GSoC (last year) regarding `scala.meta` because it requires such intricate knowledge of macros and scalac internals. I would double or triple the estimate if it's not Olaf.

heathermiller: I don't disagree that Olaf would be a good fit for this project. But again, separation of concerns and proper protocol. Management of the SC and resourcing is not the job of the proposer. Thus, it's not appropriate to require specific SC engineers to work on a specific proposed project in an SC proposal.

olafurpg: I don't think we need to mention names here. The SC has qualified people to work on this proposal other than myself. Just to take one example, @Duhemm worked on [macro tracker](https://github.com/scalamacros/macrotracker) and [parsermacros](https://github.com/Duhemm/parsermacros).

heathermiller: Again, no need to have this discussion here/now. Resourcing is not to be done in project proposals; this is the job of the SC to do conditional upon acceptance of the project proposal at the upcoming AB meeting. 😉 Let's get back to the discussion of the scope of this project.

fommil: OK, it's your process. Personally I'd never accept a timescale estimate from somebody who is giving me a requirement, and I wouldn't estimate it as a developer without knowing who was doing the work.

proposals/014-production-ready-scalamacros.md
andyscott: Would it make more sense to remove this item, or qualify with more detailed specifics?

proposals/014-production-ready-scalamacros.md
andyscott: Aah yes. They were lost in the copy-paste from the original Google doc.

andyscott: Good catch. I'll stick with using Dotty as the primary name since it's the most commonly used name at this time.

proposals/014-production-ready-scalamacros.md
olafurpg: I think we can remove this particular detail since it's rare to pattern match against these structures.

andyscott:
I believe I've addressed all the feedback. If there's anything else to add, be sure to let me know before the meeting.

fommil:
Except to say that I think `scala.meta` is perhaps the single most important future technology in IDE tooling (one ring to bind them), I have nothing more to add :smile:

adriaanm:
Very happy to see this proposal. I do have my reservations about introducing this in the 2.12 series due to the implied maintenance burden.

PR #33, opened 2017-09-19
keep the minutes in one place only

SethTisue:
we want them on the Scala Center site, and if they're there, we
shouldn't duplicate them here


heathermiller:
👍 True! Thanks Seth!

SethTisue:
I should carry this through to the old minutes, too: #34

PR #37, opened 2018-02-17
SCP-16: Accessible Scala

fommil:
To be proposed by @bvenners and @larsrh (or whoever wishes to champion this).

CC // @ragb @olafurpg @xeno-by

### DIFF
--- a/proposals/verbal-descriptions.md
+++ b/proposals/verbal-descriptions.md
@@ -126,8 +126,15 @@ Future work may include integration with the compiler reporter, REPL and
 scaladocs, along with further improvements and personalisations / localisations
 of the verbal descriptions.
 
-On the IDE / editor integration side, the tool should be easily integrated in a general porpose IDE or editor, in such a way that leverages existing accessibility support with minimal or no effort, except for user specific configuration or personalization.
+Summaries could potentially be used in text editors and IDEs for sighted
+developers much as "import folding" and "comment folding" features can hide
+boilerplate. A method definition could be folded under a clickable button with
+an appropriate description, as produced by this tool.
 
+The tool should be easily integrated in a general purpose editor in such a way
+that leverages existing accessibility support with minimal or no effort. In
+Emacs, this could be achieved with a simple function call to the CLI tool,
+piping the results to a buffer that emacsspeak is aware of.
 
 ## Cost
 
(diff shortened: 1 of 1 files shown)


propensive:
I've got a 👍 from @larsrh, so I'm taking that as a willingness to present this proposal during the meeting. (As I will be in the meeting anyway, and was involved in the preparation of the proposal, I'm happy to have Lars delegate back to me, as necessary.)

MaxwellBo:
Hi folks! I hacked out a little proof of concept in a couple of hours (https://github.com/MaxwellBo/scala-verbal-descriptions). It isn't Scala Native because scalameta's Native support isn't quite there yet, but it does handle the examples somewhat reasonably.

Input:

```
object Main {

  type ErrorsOr[A] = ValidatedNel[String, A]
  // type alias ErrorsOr type A equal to ValidatedNel of String and type A

  def foo[A: Wibble](s: String, b: Wobble[A]): Int = {
    val a = 5
    // val a with value 5
    a
  }
}

trait Monad[F[_]] extends Applicative[F] {
}
// trait Monad with higher type F extending applicative of F with five declarations

case class Person(name: String, address: Address, age: Int)
```

Output:

```
 object Main     has 2 declarations:   type alias ErrorsOr type  A     equal to ValidatedNel of String,  and A,  and
 definition foo with  A bounded by Wibble    and params  s String ,  and  b Wobble of A  with return type Int  with body 'a'  with value 5; and a
 trait Monad has type params  F       extending Applicative of F
case class Person        contains  name String ,  address Address ,  and  age Int
```

I've done something like this [for Python before](https://github.com/MaxwellBo/neoreader/blob/master/rplugin/python3/neoreader/py_ast.py), so I'd love to continue to work on this.

olafurpg:
Very cool @MaxwellBo ! 

We are close to adding native support in Scalameta. We already publish a part of our build with native but are currently waiting for the next Scala Native release to unblock a couple issues discovered in https://github.com/scalameta/scalameta/pull/1263

fommil:
@MaxwellBo thanks for taking the time to put this together!

I don't think scala-native is strictly necessary - I've had good experience with nailgun - but this does seem to be an excellent application of the technology so I can't blame anybody for wanting to do it that way :smile:

MaxwellBo:
@olafurpg Apologies, I seem to be out of the loop w.r.t Scalameta! If time permits I can see what I can do about migrating it

@fommil Looking at my calendar, I think I may still need a month. Furthermore, I think I may be unsure of the advisory board's processes - what do you mean by "increas[ing] the chances of it being accepted"?

propensive:
@MaxwellBo The proposal is for the Scala Center Advisory Board to accept or reject as a "recommendation" to the Scala Center, with the expectation that they will find a way to implement it. I think it would be great if that implementation could build upon the work you've done, or if they could support you towards completing it in some way!

PR #38, opened 2018-03-12
Add SCP-017 "Scala Center Support for LSP and STP Working Groups"

xeno-by:


proposals/017-lsp-stp-wg-support.md
```diff
@@ -0,0 +1,35 @@
...
+The Scala Center should devote developer time and resources to supporting
+the work of the LSP-WG and STP-WG in kickstarting Scala's tooling ecosystem,
+through contributions which ensure a dependable level of quality of documentation,
+test coverage and reliability of key projects identified by or arising from the
+working groups.
+
+## Costs
+
+There should be no direct costs associated with this proposal, however
+contracting work to external developers (e.g. through student projects) may be
```
heathermiller: Organizationally, contracting is impossible. We tried to do it, got in trouble with EPFL.

So I would suggest cost be a little better worked out given the current organizational structure of the SC (e.g., the developers we currently have) before this is merged.

proposals/017-lsp-stp-wg-support.md
```diff
@@ -0,0 +1,35 @@
...
+the most efficient way of achieving certain goals of the working groups.
+
+## Timescales
+
+The required input should be 2-5 developer-days per week, split between
+the two working groups, and dependent on developer availability, for one year.
```
heathermiller: So 0.5 FTE for 1 year?

heathermiller: Or you mean 0.4-1 FTE for 1 year (assuming you mean 2 to 5 days per week)

propensive: Do you have an idea about how viable this commitment (0.4-1 FTE) would be for the SC, @heathermiller? It might make sense to be more specific in the proposal, if it's known what resources are available. (Though we can always refine it during the meeting.)

heathermiller: I think the proposal should include an idea of how much engineering effort it may require. I'm just trying to determine what that number is.

proposals/017-lsp-stp-wg-support.md
xeno-by: Thank you for explaining. I have removed the contracting part from the proposal.

proposals/017-lsp-stp-wg-support.md
xeno-by: Updated the proposal to provide an exact number rather than a range.

PR #45, opened 2018-11-26
SCP-029: sbt dependency conflicts management

jto:


alexarchambault:
Related: https://github.com/coursier/coursier/issues/349

julienrf:
Something is not clear to me: how do we reliably detect conflicts? I might be wrong but my understanding is that the current eviction process is based on revision names, but nothing enforces library authors to follow a specific binary compatibility scheme in their naming.

jto:
That's a good question :) The default eviction process is indeed based on revision names. While it's not ideal, it mostly work bc. most libraries actually use semantic versioning.   

[`ConflictManager`](https://www.scala-sbt.org/1.x/api/sbt/librarymanagement/ConflictManager$.html) should let you customize the behaviour but actually does not.

propensive:
Merge pending, awaiting confirmation from @sjrd of Spotify's accession to the advisory board.

stuhood:
@alexarchambault 's point is relevant.

It would be good for the proposal to discuss whether this is something that is better addressed in the resolver (ivy/coursier).

SethTisue:
> It would be good for the proposal to discuss whether this is something that is better addressed in the resolver (ivy/coursier)

relevant: https://github.com/sbt/sbt/pull/4430 (which probably the commenters on this ticket are aware of, but maybe not everyone else knows) — if it were to move forward, then any Ivy-specific work might decrease in appeal/urgency

jto:
Hi @alexarchambault, @stuhood and @SethTisue. Thanks for the feedback and pointers :)

IMHO `ConflictManager` should at least be fully supported when SBT is used with its default configuration. 

If coursier is shipped with SBT and it is possible to replace the default resolver by it with a simple configuration, and assuming that coursier fully support all possible `ConflictManager` options, then yes the appeal of fixing the Ivy resolver decreases but:

- SBT behaviour would still be broken out of the box. Given the number of criticisms about SBT being unintuitive, I would rather avoid that.
- The process of resolving conflicts would still need to be documented.

> Related: coursier/coursier#349

`ConflictManager.strict`  does work (with Ivy at least), but it also has pretty serious drawbacks:

1. When a conflict on library `A` is detected, the user has to force the version of `A` using `dependencyOverrides`. If later in time a newer version of `A` is required (for example bc. a new dependency `B` that depends on `A` is introduced), the "old" version of `A` will still have priority which may introduce a runtime issue *again*. Over time, it becomes hard to maintains a clean dependency tree.
1. When a specific version of `A` is forced, it changes the dependency tree (because `A` is also pulling dependencies transitively) which in turn creates new conflicts. As a consequence you have to override the version of  library `B`, which creates new conflicts... Rinse and repeat. Even if you go through the tedious process of fixing all conflicts manually, upgrading any dependency or introducing a new dependency will force you to do that work all over again because it also changes the dependency tree (see point 1.).

> It would be good for the proposal to discuss whether this is something that is better addressed in the resolver (ivy/coursier).

I would rather let the implementor decide where and how a fix should be implemented. I think she is in a better position than the author to decide how a particular recommendation should be implemented. 

My understanding is that the goal of the recommendation is to set a reachable objective and to plan the work, not to make technical decision (Unless of course the author of a recommandation and the implementor are in fact the same person).

I will add those points in the proposal to keep all the context in one place.

alexarchambault:
About [point 3 of the proposal](https://github.com/scalacenter/advisoryboard/blob/master/proposals/020-sbt-transitive-dependencies-conflicts.md#proposal), what's the relation of[missinglink](https://github.com/spotify/missinglink) to that? Did you run into particular issues / blockers with it?

alexarchambault:
@jto About point 3 of the proposal, what's the relation ofmissinglink to that? Did you run into particular issues / blockers with it?

PR #54, opened 2019-11-30
Zinc improvements proposal

eed3si9n:


proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,104 @@
...
+
+As we make changes to Zinc either as bug patches or in a more major way in the future, we should make sure to bring the various downstream build tools along so they do not lag behind. The proposed task is to send pull requests to the downstream build tools.
+
+### Standardization of signature and outline types
+
+In the Scala tooling ecosystem, the notion of outline types is emerging as a key technique. In a nutshell, outline types are public signatures of a given code base scraped without going through the full typer (missing return types, super etc might still need to be resolved).
+
+Here are some of the cases for signature information (also known as API info):
+
+- Company-wide code browsing (Metadoc in Twitter).
```
olafurpg: This line can be removed. Outline types are not used for company wide code browsing at Twitter and “Metadoc” has been renamed to Scalameta/Metabrowse (and this project has very little usage)

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,104 @@
...
+- Build pipelining by starting the subproject builds earlier. (outline typing allows further parallelism)
+
+Today, the public signature information is represented as various forms:
+
+- Semanticdb
+- Scala pickle in `*.class` file
+- Zinc API info
+- TASTY
+
+Given the utility of signature information, we should standardize the representation of the signature information across Scala 2.13 and Scala 3.x and provide it via Zinc. Depending on the use case, different representation may have different degrees of details and/or performance characteristics for some operations.
```
julienrf:
```suggestion
Given the utility of signature information, we should standardize the representation of the signature information across Scala 2.13 and Scala 3.x and provide it via Zinc. Depending on the use case, different representations may have different degrees of details and/or performance characteristics for some operations.
```

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,104 @@
...
+- sbt
+- Bloop
+- Gradle via Scala plugin
+- Maven via scala-maven-plugin
+- Mill
+- Pants
+
+As we make changes to Zinc either as bug patches or in a more major way in the future, we should make sure to bring the various downstream build tools along so they do not lag behind. The proposed task is to send pull requests to the downstream build tools.
+
+### Standardization of signature and outline types
```
julienrf: This part of the proposal looks similar to [SCP-018](https://github.com/scalacenter/advisoryboard/blob/master/proposals/018-converging-214-30.md). Maybe you could mention how it is related (if it is)?

proposals/021-zinc-improvements.md
eed3si9n: a6f5dfb32b2501623ac0968b8c1f1d7a9a14af97

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,104 @@
...
+As the ecosystem prepares for Scala 3.x, we should ensure that Zinc continues to function with the new compiler.
+
+## Proposal
+
+1. Fixing under-compilations
+2. Propagating of Zinc to build tools
+3. Standardization of signature and outline types
+4. Understanding Scala 3 impact to Zinc
+
+### Fixing under-compilations
```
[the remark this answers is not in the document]
eed3si9n: Sorry, it's a word that I made up and use frequently. I've added a definition here - https://github.com/scalacenter/advisoryboard/pull/54/commits/b46774c486dc7950ad195b6f457f2f133c196126

proposals/021-zinc-improvements.md
eed3si9n: Here's my take on SCP-018 vs this. /cc @adriaanm 
https://github.com/scalacenter/advisoryboard/pull/54/commits/b3814bbec9ff844f5b4afa0cec38b55ec802674f

adriaanm:
👍 Since proposals are advisory, we felt it was better to provide a broader selection for the Scala Center to choose from. We can certainly discuss priorities at the meeting and update the proposal accordingly. We at Lightbend also intend to pick up some of the work items, as mentioned in the proposal.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,116 @@
...
+Here are some goals for the standard signature info:
+
+- Represent normal (post-typer) public signature info for Scala 2.13.
+- Represent normal (post-typer) public signature info for Scala 3.x.
+- Represent outline type (pre-typer) public signature info for Scala 2.13.
+- Represent outline type (pre-typer) public signature info for  Scala 3.x.
+- Support deserialization back into the symbol table for separate compilation in Scala 2.13.
+- Support deserialization back into the symbol table for separate compilation in Scala 3.x.
+
+Side note: [SCP-018][18] proposed that Scala 2.14 and 3.0 both emit TASTY post-typer, "so that we may have one post-type checker compiler pipeline for both versions." This proposal is less ambitious in this aspect since it targets current 2.13 and 3.x series each having their own backends. It is similar in a sense that it calls for a standard signature representation that is able to express both Scala 2.13 type system and Scala 3.x type system, except here we do not specify the representation to be TASTY. The goal of 021 is to share tooling logic for separate compilation, incremental compilation, etc.
```
stuhood: > It is similar in a sense that it calls for a standard signature representation that is able to express both Scala 2.13 type system and Scala 3.x type system, except here we do not specify the representation to be TASTY. The goal of 021 is to share tooling logic for separate compilation, incremental compilation, etc.

So, I think that a very important distinction to be drawn is between the information needed to convert an outline back into a symbol table for separate compilation (the last two items in your list above), and the information needed for incremental compilation.

As far as I know, incremental compilation does not require even remotely the level of detail that an outline-for-separate-compilation would (it probably roughly requires a "hash per symbol name"?). So while being able to more easily support incremental compilation across multiple compiler versions is a very good goal that I think that this proposal should push, standardization of outlines feels like a much lower priority: in Twitter's usage, we would probably be more than happy to say that "you can't use pipelined-compilation/outlining across a language version boundary".

As much as my feedback in the last reading of this doc was to look for a connection to outlining, having seen the result of connecting them, I would suggest narrowing the goal here to more precisely describe what is necessary for incremental compilation.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,116 @@
...
+- Wildcard imports
+- Macros
+- Implicit search
+- Default arguments
+- Name shadowing
+- Compile-time contants [zinc#227][227]
+
+Given known under-compilation issues in Zinc with regards to the above problem areas:
+
+- Could we reuse the name hashing heuristics to invalidate code arising from these under-compilation scenarios?
```
stuhood: These bullets feel disconnected from the rest of the proposal.

I would personally hope that given a deeper connection between incremental compilation and the compiler (perhaps by having the compiler optionally emit per-symbol/class digests, or something), under-compilation would be reduced/removed as a sideeffect.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,116 @@
...
+
+Some of the outlined tasks may be implemented by Lightbend Scala team directly, but in collaboration with Scala Center engineers, and using the Center as a coordination point for community and stakeholder interests.
+
+Zinc started its life as a module in sbt, outside of Scala. Closer coordination between the Scala compilers (Scala 2.x "nsc", dotc, Twitter rsc) becomes more important to respond to growing demand for better performance (see also SCP-015) and different use cases such as build caching.
+
+As the ecosystem prepares for Scala 3.x, we should ensure that Zinc continues to function with the new compiler.
+
+## Proposal
+
+1. Fixing under-compilations
```
stuhood: At a fundamental level, I would suggest reordering and editing this list into something like:
```
1. Understanding Scala 3 impact to Zinc and report on incremental compilation prior art (rustc, clangd, etc)
2. Design and implement a minimal, standardized "incremental compile metadata" format that Scala compilers (including Scala 3) can emit
3. Consume standardized "incremental compile metadata" format in Zinc
```

My feeling is that this list better accomplishes the goal set out in the abstract ("We think the next step in Zinc's evolution is to bring its compiler-specific logic into the Scala compiler itself, so that incremental compilation becomes a supported feature of the language, rather than a feature that's tacked on externally."), and would accomplish _as a sideeffect_ a reduction of under-compilation.

proposals/021-zinc-improvements.md
eed3si9n: The source-to-source dependency phase brought into the compiler (Scala 2.x or 3.x) would still be based on that of Zinc, so unless we explicitly test for the interactions of incremental compilation with each language feature, I don't think they would be fixed. Also some interactions are more tractable to deal with (compile-time constant) than others.

proposals/021-zinc-improvements.md
eed3si9n: If the goal was to only care about the incremental compilation among the same Scala version, then Zinc's API info would be good enough. I am interested in standardization this partly so we can include the signature info into the JAR. From there we could use it for incremental compilation, or use it to query things.

If we only cared about Scala 2.12 and 2.13, compiler pickle might do the job. If Zinc used compiler pickle that would remove the work Zinc extracting the API info.

The other angles to consider are use cases that's on the horizon such as pipelined-compilation, `2-depends-on-3`, and `3-depends-on-2`. If the eventual goal of Scala 2.x is to comprehend TASTY and perform separate compilation, then it might as well acquire the capability to export its types as TASTY?

lrytz: > I am interested in standardization of this partly so we can include the signature info into the JAR. From there we could use it for incremental compilation, or use it to query things.

Besides API information, incremental compilation also needs dependency data, i.e., what entitites a source file depends on (class files from same project, other projects, external libraries, inherited classes), and the "used names" (class `C` uses names `{Predef, Object, someMethod, ...}`). Do you have ideas what to do about this meta-data?

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,116 @@
+# SCP-021: Zinc improvements
+
+## Proposer
+
+Proposed by Eugene Yokota, Lightbend, November 2019
+
+## Abstract
+
+We propose enhancements to the core of the compilation toolchain, Zinc. (Zinc provides Scala compilation services, including incremental compilation, to build tools and IDEs.)  We think the next step in Zinc's evolution is to bring its compiler-specific logic into the Scala compiler itself, so that incremental compilation becomes a supported feature of the language, rather than a feature that's tacked on externally.
```
lrytz: > incremental compilation becomes a supported feature of the language

How far do yo think this should go? Moving the code that accesses compiler internals into the compiler improves maintainability, and comes with a protocol between the compiler and zinc (https://github.com/scala/scala/pull/8531), so that seems beneficial. However, it makes sense to me to keep the invalidation heuristics and the logic to invoke the compiler outside of the compiler itself.

proposals/021-zinc-improvements.md
eed3si9n: > Besides API information, incremental compilation also needs dependency data, i.e., what entitites a source file depends on, and the "used names"..

There are two levels of "we could use it for incremental compilation". Suppose there's an app A that depends on library B.

### A given B's change

First is the incremental compilation of app A given binary changes in library B, which requires only API info of B. By allowing Zinc to read this from B's JAR, my hope is that we could achieve higher granularity of incremental compilation off of any library out of Maven Central not just other subprojects.

### A given A's change

Second level is the incremental compilation of app A given the changes in A, which requires full Analysis (API info, used name etc).

For some organization interested in build caching, the idea of including full Analysis (API info, used name etc) might still be interesting, which would allow resumption of compilation off of someone else's machine. But that could be an optional flag since that would only benefit people who are doing that, and it would be dependent on Zinc version.

proposals/021-zinc-improvements.md
eed3si9n: My concern here is more about what the user experience as "Scala." Today, the experience is primarily via IDEs and build tools that uses Zinc. By upstreaming things like incremental compilation more into scalac proper, I am hoping to see better performance/correctness and also creative love and care (like `Vector` gets).

There are pros and cons to strongly tying incremental compilation to the release cadence of Scala since we may be bound to binary contract (see also `Vector`), but I'm ok to say "to get better incremental feature use latest Scala version."

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,113 @@
...
+2. During the second compilation onwards, initial sources are invalidated based on the timestamp change (or the file watcher); and Zinc tries to calculate the minimal set of transitively invalidated sources based on the relationship analysis and heuristics.
+
+Invalidating more sources than necessary would lead to unnecessary work, called *over-compilation*; invalidating fewer source than necessary could lead to invalid programs, and we call them *under-compilation*. There are known pain points in Scala 2.x in terms of under-compilation:
+
+- Package objects
+- Wildcard imports
+- Macros
+- Implicit search
+- Default arguments
+- Name shadowing
```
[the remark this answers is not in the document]
eed3si9n: Fixed in d46d6b4

proposals/021-zinc-improvements.md
dwijnand: If newer compilers maintain a stable API and own their own incremental compilation heuristics and logic (which I agree with), what's left in Zinc, apart from the heuristics and logic for older compilers?

jvican: > However, it makes sense to me to keep the invalidation heuristics and the logic to invoke the compiler outside of the compiler itself.

@lrytz It does make sense to do that IMO because Zinc is not the Scala compiler. It's an IR on top of Scala and Java compiler information *across different versions* as well as the APIs that all build tools use to interface with them.

It's important they are not confused and put together in the same bucket -- they should be evolved in different repositories with different lifecycles. In fact, it's the only thing that makes sense, as the compilers for Scala 2 and Scala 3 live in different repositories and Zinc needs to support both. Same argument for Javac.

From the suggestions I've read so far, it looks like this is a proposal to completely rewrite Zinc from scratch. I think that's not a good use of our time and we should instead limit the scope of this effort a lot more. I propose we are more pragmatic, make incremental improvements to Zinc and avoid elegance-driven designs.

I applaud the effort of moving the Scala 2 bridge to Zinc and come up with a stable API to extract the information required by both Scala 2 and Scala 3. But I'm becoming really concerned with some of the comments in this document.

> By upstreaming things like incremental compilation more into scalac proper, I am hoping to see better performance/correctness

This has been happening outside of the compiler so far, I don't think it's a good argument for upstreaming it from its repository.

proposals/021-zinc-improvements.md
jvican: > By allowing Zinc to read this from B's JAR, my hope is that we could achieve higher granularity of incremental compilation off of any library out of Maven Central not just other subprojects.

What's the pragmatic benefit of this? The fact that Zinc doesn't have support for this is not a pain point for users, in fact the invalidation works just fine without this information.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,116 @@
...
+Here are some of the build tools that use Zinc:
+
+- sbt
+- Bloop
+- Gradle via Scala plugin
+- Maven via scala-maven-plugin
+- Mill
+- Pants
+
+As we make changes to Zinc either as bug patches or in a more major way in the future, we should make sure to bring the various downstream build tools along so they do not lag behind. The proposed task is to send pull requests to the downstream build tools.
```
jvican: Is this an effort that will be done solely by the Scala Center? Or will everyone involved in the redesign have have  "skin in the game" and be involved in porting the existing use of Zinc in these build tools to the new APIs?

eed3si9n: As I wrote in the abstract it would likely be a mix:

> Some of the outlined tasks may be implemented by Lightbend Scala team directly, but in collaboration with Scala Center engineers, and using the Center as a coordination point for community and stakeholder interests.

As this is an advisory board proposal, I can only propose tasks as a stakeholder. How much resource would be  allocated I guess would be determined in the meetings?

julienrf: > How much resource would be allocated I guess would be determined in the meetings?

Yes, but we first need to have an estimation of the cost of the proposal.

proposals/021-zinc-improvements.md
eed3si9n: The higher granularity becomes useful if there's a frequent change to core library published internally to Artifactory etc, and there are lots of apps hanging off of it.

Another pragmatic benefit is compilation time speedup (by removing API extraction phase) - at least for Scala 3.x if we chose TASTY.

proposals/021-zinc-improvements.md
eed3si9n: > This has been happening outside of the compiler so far, I don't think it's a good argument for upstreaming it from its repository.

When a new language feature is proposed, historically Zinc has not been tested or updated in tandem beyond full compilation that gets exercised via Community Build. From the point of view of Scala users, a bug in incremental compilation is a bug in Scala. The more we could break down the silo of scala/scala vs Zinc, the more it would be beneficial to both the users and contributors IMO.

jvican: > When a new language feature is proposed, historically Zinc has not been tested or updated in tandem beyond full compilation that gets exercised via Community Build. 

That's because no one working on the compiler has been assigned to do that. It's a social problem, not a technical one, and it looks like this problem will be gone anyway because both Guillaume and you, as well as the engineers taking on this proposal, will be doing just that.

proposals/021-zinc-improvements.md
jvican: > The higher granularity becomes useful if there's a frequent change to core library published internally to Artifactory etc, and there are lots of apps hanging off of it.

My point is that Zinc is already successfully doing this and its approximation is good enough that no users have complained about it. Of all things we could be improving in Zinc, I'm not sure this is a use case we should be optimizing for.
  
> Another pragmatic benefit is compilation time speedup (by removing API extraction phase) - at least for Scala 3.x if we chose TASTY.

I'm not sure I understand. Invalidations of binary dependencies work at the class-file level and don't require compilation, so there's no cost in running the API phase.

stuhood: > > I am interested in standardization of this partly so we can include the signature info into the JAR. From there we could use it for incremental compilation, or use it to query things.
> 
> Besides API information, incremental compilation also needs dependency data, i.e., what entitites a source file depends on (class files from same project, other projects, external libraries, inherited classes), and the "used names" (class `C` uses names `{Predef, Object, someMethod, ...}`). Do you have ideas what to do about this meta-data?

To reiterate @lrytz's point: is the idea behind using TASTY here that we think it will _also_ give us (fully qualified?) dependency information? If so, that's potentially interesting. But TASTY is a very large component/blocker to rely on for a relatively small feature.

It's worth noting that it is _almost_ possible currently to use the built-in JVM tool `jdeps` to dump out this dependency information. I say "almost" because I'm sure that there is a little bit more information in pickles that it misses. If TASTY generically allows for the equivalent of an `sdeps`, that could be interesting.

proposals/021-zinc-improvements.md
stuhood: I view "moving something into the compiler(s)" as meaning that each compiler would need an independent test suite of that produced metadata, independent of zinc's consumption of that metadata.

So, yes: the problem does still need to be solved. But my point about improvements being a "sideeffect" was that testing the accuracy of the metadata from within a compiler's test suite seems significantly easier than outside.

proposals/021-zinc-improvements.md
stuhood: @jvican : On one hand, I agree that this proposal could be shrunk/refocused (see https://github.com/scalacenter/advisoryboard/pull/54#discussion_r352905153). But on the other hand, I completely agree with Eugene's point here: https://github.com/scalacenter/advisoryboard/pull/54#discussion_r353472072

I think that "emitting API/dependency metadata" from the compiler, in a location where it is easier to test for accuracy as part of the compiler's testsuite would be a significant step forward.

proposals/021-zinc-improvements.md
stuhood: IMO, this should not be in scope. Or should be at the end.

proposals/021-zinc-improvements.md
eed3si9n: > is the idea behind using TASTY here that we think it will also give us (fully qualified?) dependency information? If so, that's potentially interesting.

For this proposal, I am mainly concerned about the signature info, given which, downstream modules can perform incremental compilation or (normal or early) separate compilation. 

For that matter, a potential hybrid solution may be that 2.x uses pickles, and 3.x uses TASTY, and extend _both_ to support information needed by Zinc and sbt (like the ability to query for test suites).

eed3si9n: @jvican 

> I'm not sure I understand. Invalidations of binary dependencies work at the class-file level and don't require compilation, so there's no cost in running the API phase.

I should add some more context to the text then. API extraction takes up 5% of compilation last I checked. That's basically duplicated data structure created mostly because Scalac and Zinc do not cooperate today. That's the base problem I would like to address by picking a standard signature info.

proposals/021-zinc-improvements.md
eed3si9n: Speaking of shrinking this proposal a bit, the first I would chop off would be the "Understanding Scala 3 impact to Zinc." I still would like the study to happen, but that one might be the least concrete item, and perhaps it's better to come up with concrete tests/fixes (e.g. for under compilation), and let the understanding take place as a side effect.

proposals/021-zinc-improvements.md
Duhemm: Are you all speaking about the same thing? From my understanding, the proposal suggests moving only the compiler bridge inside scalac:

> We think the next step in Zinc's evolution is to bring its compiler-specific logic into the Scala compiler itself

Moving the compiler bridge to scalac is a worthwhile effort in my opinion. Dotty has been doing it for a long time now without major issues. The way I see it, the Scala team, Zinc and their users would benefit from it:

 - For scalac, this means it'd be easier to verify the impact that a compiler change has on incremental compilation
 - For zinc, this means a simpler build, simpler sources, and simpler versioning of the compiler bridge.
 - For users, this means the guarantee that their compiler has been thoroughly tested with their incremental compiler.

I feel like moving the compiler bridge inside the compiler was already discussed a long time ago, but was rejected at the time because incremental compilation and name hashing were not as stable as they are today. Since today saying "upgrade Scala to get better incremental compilation" is acceptable, there's nothing to stop us from doing it. I believe that @adriaanm worked on such a prototype already.

I agree that trying to move Zinc as a whole inside scalac is not the best use of resources. It'd indeed amount to a complete rewrite, and wouldn't bring any benefits in my opinion (harder to reuse, tied to scalac's release schedule, scalac would need to taught about new concepts such as "project", etc.) I don't think this is what's being proposed here, though.

Can we clarify in the document which new components would be added to scalac?

adriaanm: @stuhood said it well, and this is how I also understood the intent behind our proposal:
> I think that "emitting API/dependency metadata" from the compiler(s), in a location where it is easier to test for accuracy as part of the compiler's testsuite would be a significant step forward.

I don't think we should go beyond this at this point

proposals/021-zinc-improvements.md
jvican: > API extraction takes up 5% of compilation last I checked.

That doesn't match my experience, it can take from 1% to 5% in various projects, depending on different factors.
  
> I should add some more context to the text then.

I understood you were talking of API extraction in the context of binary dependencies. In that context, the cost of API extraction doesn't exist and that's what I was referring to in my previous reply.

Here's the text that made me think that:

> First is the incremental compilation of app A given binary changes in library B, which requires only API info of B. By allowing Zinc to read this from B's JAR, my hope is that we could achieve higher granularity of incremental compilation off of any library out of Maven Central not just other subprojects.

proposals/021-zinc-improvements.md
jvican: I'm glad we're all on the same page and that's the intent of the document! I understood something else while reading the comments above.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+Incremental build twice without source changes to measure the overhead of no-op. This is relevant because Zinc performs classpath lookup for all library-associated classes to validate all library JARs.
+Incremental build with source changes (add classes, add methods, change methods etc).
+Multiple modules "core" and "app" with minimum library dependencies:
+Comparison of clean build using sbt vs scalac to measure the overhead of added phases.
+Incremental build twice without source changes to measure the overhead of no-op.
+Incremental build with source changes to "core" (add classes, add methods, change methods etc).
+Two single modules "core" and "app" that uses library dependencies to depend on "core":
+Comparison of clean build using sbt vs scalac to measure the overhead of added phases.
+Incremental build twice without source changes to measure the overhead of no-op.
+Incremental build with source changes to "core" (add classes, add methods, change methods etc) to emulate binary change.
```
lrytz: The formatting seems to be missing here

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+3. Standardization of signature and outline types
+3. Propagating of Zinc to build tools
+
+### Benchmark incremental compilation
+
+We propose to add incremental compilation performance to the current compiler benchmark to quantify performance improvements.
+
+Suggested tests:
+A single module with minimum library dependencies:
+Comparison of clean build using sbt vs scalac to measure the overhead of added phases.
```
jvican: If the purpose of these benchmarks is to quantify the overhead of the additional Zinc phases, benchmarking bloop makes more sense -- it's a minimal build tool that wraps up Zinc APIs and compiles faster than sbt.

If you still want to keep sbt in the benchmarks, it's fine by me, but we should add Bloop too.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+- Build pipelining by starting the subproject builds earlier. (outline typing allows further parallelism)
+
+Today, the public signature information is represented as various forms:
+
+- Semanticdb
+- Scala pickle in `*.class` file
+- Zinc API info
+- TASTY
+
+During compilation, Zinc creates API info by adding _Extract API_ phase, which takes up approximately 5% of the compilation time. Although the information is useful to sbt (for both incremental compilation and reflective detection), there's an overlap of data with pickles / TASTY that seems wasteful.
```
jvican: See comment regarding the 5% above.

proposals/021-zinc-improvements.md
eed3si9n: Fixed it in de0ac52

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+2. During the second compilation onwards, initial sources are invalidated based on the timestamp change (or the file watcher); and Zinc tries to calculate the minimal set of transitively invalidated sources based on the relationship analysis and heuristics.
+
+Invalidating more sources than necessary leads to unnecessary work, called *over-compilation*; invalidating fewer source than necessary can lead to invalid programs, and we call them *under-compilation*.
+
+There are certain language features that leads to gotcha situation in incremental compilation. For example, using wildcard import you might be able to trick name hashing to under compile.
+
+## Expected outcome
+
+- Improved Zinc performance
+  - By Scala compiler and Zinc cooperating, there might be opportunities for speedup and memory usage reduction. Concretely, Zinc API Extraction phase (5% of compilation time) could potentially be removed if we could reuse the information from an existing signature information.
```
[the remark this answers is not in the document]
eed3si9n: This is based on real-world observation on 5M LoC.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
+# SCP-021: Zinc improvements
+
+## Proposer
+
+Proposed by Eugene Yokota, Lightbend, November 2019
+
+## Abstract
+
+We propose enhancements to the core of the compilation toolchain, Zinc. Zinc provides Scala compilation services, including incremental compilation, to build tools and IDEs. We think the next step in Zinc's evolution is closer coordination with Scala compilers. Ultimately incremental compilation should become a supported feature of the language, similar to how we treat collections, which could entail better testing and language evolution considering the impact on incremental compilation.
```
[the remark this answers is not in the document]
SethTisue: (fwiw, I found the comparison to collections peculiar)

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+2. During the second compilation onwards, initial sources are invalidated based on the timestamp change (or the file watcher); and Zinc tries to calculate the minimal set of transitively invalidated sources based on the relationship analysis and heuristics.
+
+Invalidating more sources than necessary leads to unnecessary work, called *over-compilation*; invalidating fewer source than necessary can lead to invalid programs, and we call them *under-compilation*.
+
+There are certain language features that leads to gotcha situation in incremental compilation. For example, using wildcard import you might be able to trick name hashing to under compile.
+
+## Expected outcome
+
+- Improved Zinc performance
+  - By Scala compiler and Zinc cooperating, there might be opportunities for speedup and memory usage reduction. Concretely, Zinc API Extraction phase (approx 5% of compilation time) could potentially be removed if we could reuse the information from an existing signature information.
```
dwijnand:
```suggestion
  - By Scala compiler and Zinc cooperating, there might be opportunities for speedup and memory usage reduction. Concretely, Zinc API Extraction phase (which can be 5% of compilation time) could potentially be removed if we could reuse the information from an existing signature information.
```

proposals/021-zinc-improvements.md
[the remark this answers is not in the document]
eed3si9n: That's fair. I'd be happy to clarify that ymmv.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+1. During the first compilation, class-to-class relationships and signature info are recorded.
+2. During the second compilation onwards, initial sources are invalidated based on the timestamp change (or the file watcher); and Zinc tries to calculate the minimal set of transitively invalidated sources based on the relationship analysis and heuristics.
+
+Invalidating more sources than necessary leads to unnecessary work, called *over-compilation*; invalidating fewer source than necessary can lead to invalid programs, and we call them *under-compilation*.
+
+There are certain language features that leads to gotcha situation in incremental compilation. For example, using wildcard import you might be able to trick name hashing to under compile.
+
+## Expected outcome
+
+- Improved Zinc performance
```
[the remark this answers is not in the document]
eed3si9n: I did not mean to suggest that performance is prioritized over correctness. I flipped the order here https://github.com/scalacenter/advisoryboard/pull/54/commits/a2b8677932e2593bcf8c70634dce5b3ee2f92a4a to put correctness above, but in general there's a trade-off relationship between over- and under-compilation when we're using heuristics. This is why I'm saying that the language itself needs to acknowledge incremental compilation.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Improved correctness
+  - By bringing compiler implementation specific details into Scala compiler, we expect better maintenance and test coverage over time. This applies to Dependency tracking phase.
+  - When heuristics are used, there's a trade-off between over- and under-compilation (currently it focuses on local development). We should provide a knob to adjust this for CI usages.
+- Improved Zinc performance
+  - By Scala compiler and Zinc cooperating, there might be opportunities for speedup and memory usage reduction. Concretely, Zinc API Extraction phase (which can be 5% of compilation time; your mileage may vary) could potentially be removed if we could reuse the information from an existing signature information.
+
+## Proposal
+
+1. Benchmark incremental compilation
```
stuhood: The numbering in here renders correctly, but is a bit confusing as text.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+1. Fixing under-compilations
+3. Standardization of signature and outline types
+3. Propagating of Zinc to build tools
+
+### Benchmark incremental compilation
+
+We propose to add incremental compilation performance to the current compiler benchmark to quantify performance improvements.
```
stuhood: Does this assume that Lightbend is going to be driving the work that this depends on (ie: actually including some incremental phases in Scala 2 and Scala 3)? How long do you expect that this work will be blocked?

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Package objects
+- Wildcard imports
+- Macros
+- Implicit search
+- Default arguments
+- Name shadowing
+- Compile-time constants [zinc#227][227]
+
+The proposed task is to add tests of incremental compilation in scala/scala repository, and provide fixes where possible. In a situation where heuristics might be used consider providing "strict" mode where it biases towards correctness and over-compilation.
```
stuhood: Unless _a lot_ of zinc migrates into the compiler, this is going to be challenging to do I think. My understanding based on the first round was that the proposal was to move "dependency" and "API" metadata into the compiler, which zinc would then consume. Is that not the case?

I've lost track of what actual API boundary you are proposing here between the compiler and zinc, and it would be good to make that clearer in order to get a better idea of where various pieces will live.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+- Gradle via Scala plugin
+- Maven via scala-maven-plugin
+- Mill
+- Pants
+
+As we make changes to Zinc either as bug patches or in a more major way in the future, we should make sure to bring the various downstream build tools along so they do not lag behind. The proposed task is to send pull requests to the downstream build tools.
+
+## Cost
+
+Unknown at this stage.
```
stuhood: This should be made concrete, IMO. I can say that (on Twitter's behalf) I am unlikely to vote for an unbounded proposal. If the hesitance to add a cost is that you don't know which portions are likely to be worked on, one strategy might be roughly estimating each of your four items.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Timescales
+
+Unknown at this stage.
```
stuhood: Given that some of the proposed efforts depend on work that Lightbend are doing (afaict?), including a rough timeframe for when the Scala Center's involvement would be most helpful would be good.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Abstract
+
+We propose enhancements to the core of the compilation toolchain, Zinc. Zinc provides Scala compilation services, including incremental compilation, to build tools and IDEs. We think the next step in Zinc's evolution is closer coordination with Scala compilers. Ultimately incremental compilation should become a supported feature of the language, similar to how we treat collections, which could entail better testing and language evolution considering the impact on incremental compilation.
+
+Some of the outlined tasks may be implemented by Lightbend Scala team directly, but in collaboration with Scala Center engineers, and using the Center as a coordination point for community and stakeholder interests.
+
+Zinc started its life as a module in sbt, outside of Scala. Closer coordination between the Scala compilers becomes more important to respond to growing demand for better performance (see also SCP-015) and different use cases such as build caching.
+
+As the ecosystem prepares for Scala 3.x, we should ensure that Zinc continues to function with the new compiler.
```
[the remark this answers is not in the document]
stuhood: +1... are we now assuming that that work will be done by Lightbend or the Scala team instead?

proposals/021-zinc-improvements.md
eed3si9n: I was thinking about https://github.com/scala/compiler-benchmark / https://scala-ci.typesafe.com/grafana/dashboard/db/scala-benchmark.

proposals/021-zinc-improvements.md
eed3si9n: Fixed in https://github.com/scalacenter/advisoryboard/pull/54/commits/c3a9b14b993f8fc3cf709ec502693acc4b14e893

proposals/021-zinc-improvements.md
eed3si9n: Added links here https://github.com/scalacenter/advisoryboard/pull/54/commits/07bb6f157403440800f49872686e1aee843cf651

proposals/021-zinc-improvements.md
eed3si9n: Here's my attempt at estimating this - https://github.com/scalacenter/advisoryboard/pull/54/commits/6d57993a1b8d871a31f49f263f60eefae03cf5a9

proposals/021-zinc-improvements.md
eed3si9n: Added timing to https://github.com/scalacenter/advisoryboard/pull/54/commits/6d57993a1b8d871a31f49f263f60eefae03cf5a9

proposals/021-zinc-improvements.md
eed3si9n: I am assuming you're responding to the "add tests of incremental compilation in scala/scala repository..." part:

> My understanding based on the first round was that the proposal was to move "dependency" and "API" metadata into the compiler, which zinc would then consume. Is that not the case?

That's correct. The system under test would be the Scala compiler and the compiler bridge (that contains xsbt-dependency phase). To test for under compilation in PR validation etc, we'd depend on a compatible version of Zinc in `Test` configuration, partest, or scripted.

> I've lost track of what actual API boundary you are proposing here between the compiler and zinc, and it would be good to make that clearer in order to get a better idea of where various pieces will live.

I can add that info in the background.

eed3si9n: Here's more background - https://github.com/scalacenter/advisoryboard/pull/54/commits/4aca8f7d372cce4f83eae31d3fbe725cb01d76a9

sjrd:
I've tried keeping track of everything that's been discussed in this proposal, and waw, it's a lot.

Here is my summary of the most important things, to try and pin down my understanding of the proposal:

* Move the compiler bridge to scala/scala. This is mostly a technical change that Lightbend proposes to work on themselves. That design has been validated in dotty already.
* Make sure Zinc works across Scala 2 and 3. IMO this is the core of the proposal. @jvican identified and stressed this point already, +1'ed by @stuhood.
* Miscellaneous maintenance work: benchmarks, fixing bugs, improving performance.

From the second bullet, which is a high-level *goal*, the proposal goes into great length to discuss one design solution. That design is based on standardizing a signature info API. Somehow, most of the proposal revolves around this technical design. IMO this is where this proposal is getting side-tracked and receives a lot of debate. TBH I believe these discussions do not belong in a Scala Center Proposal to begin with. We should be having these discussions in live design meetings between the maintainers of scalac and Zinc and the engineers who are going to work on a solution. The proposal currently already embeds *one* possible technical solution to the problem space.

In my opinion, we should shrink down the content of the *proposal* per se to the high-level goals of

* ensuring that Zinc continues to work with Scala 2 and Scala 3
* work toward a design solution that achieves the above, and in the process increases our ability to implement *correct* incremental compilation for both (no under-compilation)

To that end, here is what I think we can concretely set as a realistic, actionable first task:

* 3 full-time engineer month to coordinate and participate in the design and prototyping of a solution for Zinc support across Scala versions. At least the design aspect should be done in collaboration with the relevant maintainers.

Following the results of collaboratively designing a solution that satisfies all the relevant maintainers, we can dive into more concrete implementation work, bug fixing, and performance improvements.

proposals/021-zinc-improvements.md
```diff
@@ -0,0 +1,211 @@
...
+|          |  +------------+    +--------+-----------+
+|          |                             |
+|          |                             |
+|          |                    +--------v-----------+
+|          |                    | Scala compiler     |
++----------+                    +--------------------+
+```
+
+1. Build tools communicate only with Zinc. Zinc exposes a method called [`compile(...)`][1] and everything else is opaque to the build tools.
+2. To abstract over the Scala compiler implementations Zinc talks to the compilers via compiler-interface. We are moving this out of Zinc to [sbt/compiler-interface][2].
```
jvican: Unrelated to the proposal itself, but I've just opened a ticket to move the compiler interface repo to another namespace than sbt. The repository `scala/compiler-interface` is more appropriate for this than `sbt`.

eed3si9n:
@sjrd Thanks for the feedback.

> In my opinion, we should shrink down the content of the proposal per se to the high-level goals of...

I keep hearing that the AB needs to be able to estimate the cost of proposal, so I tried to be as concrete as I can on what I think we should work on. Is this proposal considered over-specified because some members agree on the direction, but not necessarily the approach?

stuhood:
@eed3si9n : I'll wait until tomorrow afternoon to merge this, in case you'd like to make edits to respond to @sjrd's feedback. Thanks for your patience!

eed3si9n:
@stuhood Following @sjrd's feedback I made the proposal statements high level, and moved the activities into the non-binding discussion section. Thank for your patience.

PR #55, opened 2019-12-04
Proposal to complete SCP-11: implement JSR-45 support in Scala for improved debugging experience

olafurpg:


proposals/021-jsr-45.md
```diff
@@ -0,0 +1,52 @@
+# Complete SCP-11: implement JSR-45 for improved Scala debugging experience
```
julienrf: The content seems to be a subset of SCP11, rather than a superset, what do you think?

proposals/021-jsr-45.md
```diff
@@ -0,0 +1,52 @@
...
+This is a proposal to prioritize the completion of
+[SCP-11](https://github.com/scalacenter/advisoryboard/blob/master/proposals/011-debugging-symbols.md).
+To reduce the scope of SCP-11, this proposal suggests to focus only contributing
+JSR-45 support to the Scala 2 compiler.
+
+[JSR-45](https://jcp.org/en/jsr/detail?id=45) is an approved specification that
+establishes standardized tools for correlating Java byte code to source code of
+languages other than the Java programming language. For example, the Kotlin
+compiler uses JSR-45 to improve the developer experience when working with
+debugging tools and code coverage tools.
```
julienrf: Would it be possible to give a concrete example of pain caused by the absence of support for JSR-45 in Scala? I’ve been using JVM debuggers on Scala code but I’m not sure what would be improved or changed if JSR-45 was supported.

proposals/021-jsr-45.md
```diff
@@ -0,0 +1,52 @@
...
+
+Tools in the Scala community that could benefit from JSR-45 include:
+
+- IntelliJ Scala debugger.
+- Metals once the upcoming debugging support is released.
+- Code coverage tools that operate on byte code such as intellij-coverage and
```
stuhood: @olafurpg: Would this obviate the need for `scoverage`?

proposals/021-jsr-45.md
olafurpg: The discussions for SCP-11 give several concrete example use-cases:

https://github.com/scalacenter/advisoryboard/pull/21#issuecomment-290023966

That whole discussion is worth reading. My understanding is that IntelliJ already does several customizations to work around these issues, making perhaps JSR-45 less relevant.

PR #56, opened 2020-01-13
Proposal for build server protocol in sbt

jastice:
Proposal to implement Build Server Protocol support for sbt, either as an sbt plugin initially, and migrate to sbt core once stable, or directly in sbt core.

### DIFF
--- a/proposals/023-sbt-bsp.md
+++ b/proposals/023-sbt-bsp.md
@@ -1,0 +1,74 @@
+# Build Server Protocol Support for sbt
+
+## Proposer
+
+Written by Justin Kaeser, JetBrains, January 2020
+
+## Abstract
+
+The [Build Server Protocol](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md) was developed by the Scala Center with cooperation from JetBrains to ease integration of build tools with IDEs and language servers. Currently the IntelliJ Scala plugin
+and Metals rely on injecting sbt plugins into user projects to export the build structure. This approach causes friction that can be addressed by supporting the Build Server Protocol directly within sbt.
+
+## Proposal
+
+Implement Build Server Protocol support for sbt, either as an sbt plugin initially, and migrate to sbt core once stable, or directly in sbt core.
+
+### Rationale
+
+Integration of IntelliJ and Metals with sbt is burdened by some overhead and friction. The IntelliJ Scala plugin injects several sbt plugins into a user's project to be able to export the project model and interact with the sbt shell. This is quite brittle and does not give a good user experience in terms of IDE integration.
+
+Metals injects the sbt-bloop plugin to export Bloop config files, and then uses Bloop directly for compilation. Bloop already implements BSP as a server, so towards the frontend there is a good user experience. The drawback is that Bloop does not run the full sbt task graph, so anything more sophisticated than just compilation, such as source code generation or packaging, still needs to be done from sbt directly.
+
+[BSP support is on the shortlist for sbt 1.4](https://discuss.lightbend.com/t/roadmap-for-sbt-1-4/5038/9). This proposal is to ensure the resources to make it happen.
+
+### Inclusion in sbt core
+
+Adding BSP support as a plugin has the drawback that either the user needs to configure their environment or projects to explicitly install the plugin, or BSP clients need to inject the plugin into the user's sbt environment, requiring tool-specific logic.
+
+BSP is designed to enable integration between IDEs/language servers and build tools without either side needing to know specifics of each other. Native support in sbt makes this possible with the least friction for the user.
+
+### Supported endpoints and functionality
+
+An implementation should support at least:
+
+* [server lifetime endpoints](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#server-lifetime)
+* [workspace/buildTargets](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#workspace-build-targets-request)
+* [buildTarget/sources](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#build-target-sources-request)
+* [buildTarget/dependencySources](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#dependency-sources-request)
+* [buildTarget/resources](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#resources-request)
+* [buildTarget/compile](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#compile-request)
+* [Scala extensions](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#scala)
+
+An implementation with good feature coverage will also include:
+
+* [sbt extension](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#sbt)
+* [task notifications](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#task-notifications)
+    * for compilation of individual subprojects
+    * for individual sbt tasks
+* [buildTarget/test](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#test-request)
+    * [test notifications](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md#test-notifications-1) per test suite / test
+
+### BSP connection protocol
+
+BSP defines a [standard way](https://github.com/scalacenter/bsp/blob/v2.0.0-M2/docs/bsp.md#bsp-connection-protocol) for clients to discover and connect to BSP servers. One way to support this from sbt is create a BSP wrapper around an sbt shell process.
+
+### Interaction with sbt server
+
+The [sbt server protocol](https://www.scala-sbt.org/1.0/docs/sbt-server.html), like BSP, can be viewed as an extension of LSP. It could thus be a reasonable approach to extend sbt server with BSP endpoints.
+
+### Related work
+
+* [BSP support for Mill](https://github.com/lihaoyi/mill/pull/664)
+* [BSP support in Fury](https://github.com/propensive/fury/pull/297)
+* [sbt-bloop plugin](https://github.com/scalacenter/bloop/tree/master/integrations/sbt-bloop)
+* [sbt BSP plugin for Dotty proof of concept](https://github.com/dotty-staging/dotty/blob/ide-compile/sbt-dotty/src/dotty/tools/sbtplugin/DottyBSPPlugin.scala)
+
+
+## Cost & timescales
+
+* Alexandra Dima's BSP support for Mill, with almost no prior exposure to Scala or Mill, took about 2 months for a largely complete implementation
+* Justin Kaeser's BSP support in Fury implements only project structure resolution and took 1-2 weeks.
+
+I expect sbt BSP support to be more complex, but overall similar in scope.
+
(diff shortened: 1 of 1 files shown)


proposals/023-sbt-bsp.md
```diff
@@ -0,0 +1,74 @@
...
+The [Build Server Protocol](https://github.com/scalacenter/bsp/blob/master/docs/bsp.md) was developed by the Scala Center with cooperation from JetBrains to ease integration of build tools with IDEs and language servers. Currently the IntelliJ Scala plugin
+and Metals rely on injecting sbt plugins into user projects to export the build structure. This approach causes friction that can be addressed by supporting the Build Server Protocol directly within sbt.
+
+## Proposal
+
+Implement Build Server Protocol support for sbt, either as an sbt plugin initially, and migrate to sbt core once stable, or directly in sbt core.
+
+### Rationale
+
+Integration of IntelliJ and Metals with sbt is burdened by some overhead and friction. The IntelliJ Scala plugin injects several sbt plugins into a user's project to be able to export the project model and interact with the sbt shell. This is quite brittle and does not give a good user experience in terms of IDE integration.
```
julienrf: Would it be possible to provide an example of bad user experience due to that?

julienrf: Another motivation could be that the IntelliJ Scala plugin would not anymore have to inject plugins into users’ projects, and this would simplify the maintenance work of IntelliJ, right?

jastice: Bad user experience: in some environsments this fails or has unintended side effects in interaction with other plugins.

Yes, reducing maintenance load for future versions is a major motivations.

proposals/023-sbt-bsp.md
```diff
@@ -0,0 +1,74 @@
...
+
+Metals injects the sbt-bloop plugin to export Bloop config files, and then uses Bloop directly for compilation. Bloop already implements BSP as a server, so towards the frontend there is a good user experience. The drawback is that Bloop does not run the full sbt task graph, so anything more sophisticated than just compilation, such as source code generation or packaging, still needs to be done from sbt directly.
```
jvican: > Metals injects the sbt-bloop plugin to export Bloop config files, and then uses Bloop directly for compilation.
  
The next stable version of Bloop no longer works like this because of compile deduplication.

eed3si9n:
Given BSP stands for Build Server Protocol, I think it would be a good validation (both for the protocol and sbt server) that there can be a direct binding for sbt, and is usable for operations such as IntelliJ import. I would be happy to provide technical assistance if changes in sbt are needed to facilitate this goal.

As mentioned by @steinybot https://github.com/scalacenter/advisoryboard/pull/56#issuecomment-573965174, I have started looking into some of that in https://github.com/sbt/sbt/pull/5207.

jvican:
BSP stands indeed for Build Server Protocol and it's built for clients. From that point of view, the protocol has already been validated enough by clients in the Scala world, e.g. it powers both Metals and IntelliJ import use cases as well as others. Implementing it in your favorite build tool is just an implementation detail that as described above has important shortcomings.
  
I don't think pouring Scala Center resources into this effort as it is laid out here is the most effective way of bringing better tooling for end Scala developers.

smarter:
> Implementing BSP in sbt requires locking down sbt to execute every task/command so it doesn't support concurrent clients (e.g. while your terminal is running a command, Metals cannot compile your project). 

This point is critical. Any sbt/bsp integration proposal needs to have a plan for addressing the locking problem since it's so fundamental to providing a good user experience.

> A more promising approach than this proposal is not to implement BSP in sbt and instead piggyback on the existing sbt server and have Bloop, which is always running in the background, act as a broker.

Alternatively, if sbt feels like it needs more control, the reverse might be achievable: make sbt a BSP server, but have it delegate any blocking task like compile/run/... to bloop, but I have no idea how well that would work :).

I think that sbt 1.3 assimilating sbt-coursier was a huge success story: it's so great that we now have one definitive dependency resolver instead of duplicating effort working on multiple ones! It'd be great if the same convergence could be achieved with sbt-bloop.

dwijnand:
> Implementing BSP in sbt requires locking down sbt to execute every task/command so it doesn't support concurrent clients (e.g. while your terminal is running a command, Metals cannot compile your project).

This assumes that it's implemented within the bounds of sbt's command processing engine.  sbt could instead support BSP and parallel compilation using the kind of isolation that Bloop also implements.

smarter:
> sbt could instead support BSP and parallel compilation using the kind of isolation that Bloop also implements.

Indeed, but that sounds like it would require sbt to reimplement half of bloop, so why not use the existing thing instead ?

dwijnand:
It might prove to simpler to implement, maintain and distribute than either of the two sbt+bloop integrations directions.

julienrf:
I think it’s premature to discuss this level of implementation details now. It’s good to see that there are several options to implement that proposal, and if the proposal is accepted all the options will be of course considered.
But for now, please leave the _how_ aside and comment only if one of these points is unclear: the motivation (_why_), the goal (_what_), or the budget.

darjutak:
Hi Everyone!

Thank you @jastice for bringing this up, as confirmed in the thread, it is a proposal we (The Scala Center) should definitively look into. Here are a couple of things I need to state for the record:

1) Proposals to the Scala Center need to be submitted by Advisory Board members
2) The community can submit proposals through the Community Representatives (Rob Norris and/or Bill Venners)
3) The proposal should (at least) have these sections to be considered: https://github.com/scalacenter/advisoryboard/blob/master/templates/proposal.md
4) Once proposed it will be discussed by the Advisory Board members in the first coming quarterly meeting (next one in March 2020, for example), voted on and so forth.

With that being said, this discussion is very helpful for us to identify some of the important points, thank you for bringing all the sides up.

My vote will always be for long term sustainability and maintainability, stable releases and fewer layers of dependencies, even if it means more initial investment. We at the Scala Center need to think about long-term strategy too while allowing the space for incubating and experimenting with different solutions.

I now suggest that Rob and Bill take over, get the necessary information, and create a comprehensible proposal (adding basically the pros and cons on top of the original text) in order to help all of the above-mentioned to proceed smoothly.

Thanks again!

bvenners:
I'm happy to champion this proposal. As a first step, I'll connect with @jastice.

stuhood:
I agree with https://github.com/scalacenter/advisoryboard/pull/56#issuecomment-574450947 's comments. Whether it is via delegation or direct implementation, more server implementations of BSP is a Good Thing.

olafurpg:
I also agree with @smarter in https://github.com/scalacenter/advisoryboard/pull/56#issuecomment-574450947

I think it will require non-trivial effort to reimplement some core Bloop features in sbt, for example how Bloop handles concurrent compile request and isolates classpaths between clients. The IDE should not be blocked from compiling code because the user has an open REPL session. Likewise, Bloop's classpath isolation ensures that the user does not get errors like `ClassNotFoundException` in the terminal because the IDE is compiling in the background.

These features can certainly be implemented from scratch in sbt, but I'm not sure if that's a good use of Scala Center engineer's time when the functionality is already implemented in Bloop and serving users of sbt, Mill, Gradle, Maven and other build tools.

I still see huge value in having an sbt BSP server that delegates to Bloop under the hood but also triggers source/resource generators (sbt-buildinfo, protobuf, etc). When using Metals, I have several times been bitten by the fact that `sbt clean` in the terminal causes compile errors in Bloop because the generated `BuildInfo.scala` file got removed.

sjrd:
Sorry I'm very late to the party. I kept delaying answering on this thread because of Scala.js 1.0.0. :s

We've discussed this within the team at the Scala Center, and here are our aggregate conclusions regarding this proposal.

---

First, let me recap the three strategies that have been mentioned:

1. Integrate sbt-bloop into sbt core
2. Implement BSP in sbt core (probably through sbt server), directly using the task graph
3. Implement BSP in sbt core (probably through sbt server), delegating to Bloop for tasks that bloop can handle

The text of the proposal describes Strategy 2. Strategy 3 is what @smarter then @olafurpg explained. Strategy 1 is basically technologically equivalent to the status quo, but potentially with a better integrated experience.

Strategy 1 would have "minimal" work impact, and would leverage the existing tools the most. It does not solve the core issues that this proposal wants to address, though, namely the fact that task dependencies beyond what is supported by bloop are never executed across BSP commands and hence from IDEs.

Strategy 2 solves these issues, but introduces a significant problem with the locking of the sbt command pipeline, as was highlighted by @jvican.

Strategy 3 seems to offer the best of both worlds, and seems to be favored in the comments so far. It should provide the user-visible benefits of both, as well as reusing a significant amount of existing implementations (though less than Strategy 1).

---

Usually, we wouldn't discuss the particular implementation strategy for a proposal at this stage and in the AB, as it is the Center's responsibility to choose what fits best. However, in this case the choice of strategy has a significant impact on *who* would be the best person/team to implement the proposal.

In particular, we believe that Strategy 3 would be best implemented, by far, by existing core developers of sbt, namely @eed3si9n / @dwijnand. Indeed, this mostly requires deep expertise with sbt's internal workings, as well as sbt server. The required expertise in BSP itself is minimal, and is supported by several existing implementations, as the proposal lists.

If someone at the Scala Center would take this on instead, I estimate that they will need more time to get familiar with the relevant areas of sbt than to implement the proposal itself. This is clearly not an efficient way to allocate resources, *unless* we decide that it is a strategic investment for the long term that one person at the Scala Center should get familiar with sbt's inner workings. At the moment and in the team, we do not think that such a long-term investment is really appropriate.

Therefore, our overall sentiment here is that, although we wholeheartedly support the intention and outcomes of this proposal, the best path forward is for sbt core devs (or existing core contributors) to take ownership of this proposal, rather than the Scala Center.

adriaanm:
While we support this proposal, Lightbend doesn't currently have it on our roadmap for sbt, and we're not planning to change that in the near future.

sjrd:
That is unfortunate but not unexpected. In that case, I guess the options are:

* Not do anything
* Apply Strategy 1, despite its shortcomings
* Invest into long-term expertise building within the Scala Center, and do Strategy 3 anyway

jastice:
I'm hapy to support any strategy whoever actually ends up working on this takes, but here's my points on it:

1. integrating sbt-bloop into sbt core does not offer much benefit over existing tooling. Both Metals and IntelliJ (in upcoming 2020.1 release) already have Bloop-specific support for sbt to reduce import friction.
2. given the central role within Scala tooling sbt has and, for the forseeable future, will continue to have, expertise for it within the Scala Center seems like a prudent investment.
3. I'm happy to cooperate and invest some of my own time with whoever works on this.

eed3si9n:
While it's not on Lightbend's roadmap, I've been working on BSP support in my weekends. This is built on top of sbt server (https://github.com/sbt/sbt/pull/5443). Whether the actual compilation work is done by directly via sbt (Zinc) or not, hopefully the draft PR sketches out the "design patterns" of how BSP support in general can be implemented with a combination of
1. immediate reply from `NetworkChannel` and
2. asynchronous reply from a command or an input task

smarter:
> Lightbend doesn't currently have it on our roadmap for sbt

> While it's not on Lightbend's roadmap

Since it's been referred to a few times, it'd be helpful for the discussion (and other sbt-related proposals) if we could have a link to this roadmap.

stuhood:
@bvenners: If you are still interested in championing this proposal, we should make it a goal to land it in the next week or so to ensure that there is no confusion about what is being voted on on March 27th. Please work with @jastice to adjust the proposal however you see fit (based on the feedback above), and let us know when it is ready for a final round of review before merging.

bvenners:
I submitted this with a few tweaks in https://github.com/scalacenter/advisoryboard/pull/61. Can probably close this. I will add a link back to the discussion here to the new PR.

stuhood:
@bvenners : Thank you: commented there that unless anyone has further feedback (cc subscribers here!) we can merge that in the next two or three days.

PR #57, opened 2020-02-18
update attendees list

SethTisue:


SethTisue:
merging since I'm sure it's an improvement. Darja it would still be good if you double-checked that the list is now 100% correct

SethTisue:
merging since I'm sure it's an improvement. Darja it would still be good if you double-checked that the list is now 100% correct

PR #65, opened 2020-07-02
SCP-024 was accepted

SethTisue:


nmcb:
Are minutes of the 2020 meeting available ?

SethTisue:
@nmcb They're coming. I'll draft them this week and then wait a week before publishing to give the board members time to check them over and suggest any corrections.

PR #69, opened 2021-02-12
update board members

SethTisue:


SethTisue:
Unsure about the GS change. Several member companies have been sending a somewhat rotating cast of people to the meetings and I'm not always sure who the "official" representative currently is or whether it really matters. For simplicity, I suggest we list a single name per company, which is why I cut GS down to one name. (Regardless, no strong opinion about how to handle this.)

sjrd (approved):
Looks good to me as far as I can tell, but I'm also requesting a review from @darjutak.

SethTisue:
in the interest of getting this merged, I changed this to leave the Goldman entry alone.

PR #71, opened 2021-02-12
invitations and agendas

SethTisue:


SethTisue:
@adriaanm do you recall if there was agenda for the most recent meeting? I couldn't find one anywhere.

(it's not terribly important if we can't find one, since people are much more likely to want to consult the minutes and see what actually happened, rather than read about what we thought might happen...)

SethTisue:
I'm going to merge this regardless so I can link to these files from the minutes, but of course updates remain possible...

adriaanm:
There was no agenda, no.

PR #79, opened 2021-11-01
Refactoring proposal

eed3si9n:


julienrf (commented):
Thank you @eed3si9n for writing this proposal.

Overall, I think it would be easier to discuss it if it has more details, especially about the problem you want to solve, and the scope of the work to be done by the Scala Center.

For instance, when I read the proposal it seems to me that it consists of enriching BSP with endpoints for refactoring, and implementing these endpoints by calling scalafix. However, the two mentioned examples (renaming and moving) are refactorings that are not supported by Scalafix. I expect that implementing those two refactorings are not part of the proposal? Or, is it?

proposals/027-refactoring.md
```diff
@@ -0,0 +1,71 @@
...
+Proposed by Eugene Yokota, Twitter, November 1st, 2021
+
+## Abstract
+
+We propose enhancements to the tooling ecosystem around refactorings.
+
+While Scalafix has been successful in automatic rewrites, there a number of challenges:
+
+1. The rule writing is not accessible outside of the core library author persona.
+2. Refactoring often involves build changes, but Scalafix is unaware about that.
```
olafurpg, line 12: What build tools would be in scope for this project? Scalafix can edit sbt and Mill build files but not Gradle/Bazel/Maven files.

proposals/027-refactoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+Proposed by Eugene Yokota, Twitter, November 1st, 2021
+
+## Abstract
+
+We propose enhancements to the tooling ecosystem around refactorings.
+
+While Scalafix has been successful in automatic rewrites, there a number of challenges:
+
+1. The rule writing is not accessible outside of the core library author persona.
```
olafurpg, line 11: For simple rename symbol refactorings, you don't need to write a Scalafix rule. I can't remember the exact syntax but something like `rename:core.A.foo/core.A.bar` would rename the method `foo` on the class `core.A` to `bar`.

proposals/027-refactoring.md
```diff
@@ -0,0 +1,71 @@
...
+2. Refactoring often involves build changes, but Scalafix is unaware about that.
+3. Driving the Scalafix is limited to CLI or sbt.
```
olafurpg, line 13: You can also rename symbols through Metals, which uses similar SemanticDB infrastructure as Scalafix.

olafurpg (commented):
It's a bit difficult to know exactly what would be the success criteria of this proposal. I feel like there are two distinct problems:

1. We want to implement a refactoring rule that can move classes and update the build config accordingly. This feature does not exist in Metals today but it would be very nice if it was implemented.
1. Metals won't be able to apply refactoring rules for huge monorepo codebases where it's common to only import a small subset of the codebase into an IDE. It would be nice if Metals could instead serialize the refactoring rule somehow so that the build tool can run the refactoring in batch mode.

The first problem seems like a great starting point and would be a worthwhile project on its own. Problem 2 is dependent on problem 1. It's probably not too difficult for Metals to emit some kind of JSON file that describes the rename, which could then be understood by Scalafix (and buildozer-like tools).

proposals/027-refactoring.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Define Build Server Protocol for common refactoring operations.
+- Implement a reference implementation using Scalafix.
+- Implement a reference UI
+
+## Discussion
+
+Generally speaking, IDEs such as IntelliJ and Metals have intuitive user experience of refactoring operations such as renaming and moving the packages around. This makes us think that Build Server Protocol might be a good fit; however, we probably do not want to require full language indexing either so they can be used in CI and/or at the corporate monorepo scale.
+
+In Language Server Protocol, there are [Rename Request][lsp-rename] and [Code Action Request][lsp-car], which might be considered for adopting as-is. Similarly, we might take advantage of the work that's been done in [The Language Server Index Format (LSIF)][lsif] as an optional step to narrow down the refactoring targets without invoking the language server or build tools.
```
olafurpg, line 25: I don't think LSIF is necessary here, lsif-java invokes the build tool to compile the codebase with the SemanticDB compiler plugin. Any build tool that can emit SemanticDB should be sufficient.

proposals/027-refactoring.md
julienrf: I might be digressing but, does that `rename:…` command work reliably? What if there are both a class and an object named `A` with a method `foo` on it? What if the method `foo` is inherited from a parent class? Does the command update usage of the method in other files?

proposals/027-refactoring.md
eed3si9n: Once it's in the BSP spec, then I'd imagine Build Server implementations could provide part of the support, similar to buildozer for Bazel? The reference implementation could be based on sbt and Mill if those exist already.

eed3si9n:
@julienrf

> Overall, I think it would be easier to discuss it if it has more details, especially about the problem you want to solve, and the scope of the work to be done by the Scala Center.

Both in monorepo and open source, massive refactoring could enable migration of libraries and Scala versions.

Two problems are:
1. Making the initiation of refactoring as easy as IDE-based refactoring.
2. Making the application of refactoring scale over monorepo or cross-GitHub-repos.

In terms of the Scala Center's work, I think the coordination between potential backend implementations and designing specification would be a useful outcome.

> For instance, when I read the proposal it seems to me that it consists of enriching BSP with endpoints for refactoring, and implementing these endpoints by calling scalafix. However, the two mentioned examples (renaming and moving) are refactorings that are not supported by Scalafix. I expect that implementing those two refactorings are not part of the proposal? Or, is it?

It sounds like Scalafix might have rename? Even if it doesn't having some easy-mode that could generate rename (`Patch.replaceSymbols`?) rules would be a pretty good outcome I think.

proposals/027-refactoring.md
eed3si9n: So enumerating finding all references to a symbol could also be something we can lift from LSP to BSP? We basically want to avoid calling scalafix against tens of thousands of targets if we can narrow down to necessary ones.

proposals/027-refactoring.md
olafurpg: If the build server knows how to apply the refactoring, would it make sense for IntelliJ/Metals to apply the refactoring directly via BSP instead of serializing the patch to a file. 🤔

SethTisue:
not really sure if edits on this are complete, but the meeting is coming up in less than 48 hours, so I'm merging this in its current form, so people can have something to review before the meeting.

proposals/027-refactoring.md
tgodzik: Could we clarify this point? I am not sure what exactly it means. Anyone can write rule and there are examples of different rules outside scalafix main repo.

sjrd:
Here is some feedback from the Scala Center team.

Our main concern with this proposal, as it stands, is that it defines a specific *solution* without clearly stating the *problem*. Normally, Scala Center Proposals first and foremost define a problem statement, along with a possible direction for the solution. Here, it *seems* to us that the problem is that of *refactoring in the context of monorepos*. If that is indeed the case, we would like the proposal to clearly state that, and state the issues that are currently faced when trying to perform refactoring in a monorepo.

Related to the above, it is not clear that BSP would be the best solution to the above problem, which highlights why stating the problem is very important. In particular, it could be the case that the existing BSP information such as build targets and their sources/dependencies/options would be enough for a BSP *client* to implement refactorings across a monorepo. And that without having to specify more stuff inside BSP (which has its limitations, as every new refactoring would have to go through that BSP specification).

Finally, the proposal states that refactorings sometimes need to change the build definition itself, with Moving as an example. Moving can be decomposed into two independent refactorings:
1. Move the class in a different *package*, but within the same *project*. This is equivalent to the first Renaming use case.
2. Moving a class without changing its package or name to a different *project*.

The second step requires no language-level awareness, but requires to potentially change the build definition, which might impossible to do for sbt, mill, or other programmatic build tools.

In conclusion, we would like the proposal to be reviewed to clearly state the *problem* that needs solving, and only incidentally the proposed solution that is currently the main focus.

SethTisue:
There will be a bit more detail in the meeting minutes, but for now I'll briefly state that rather than vote on the proposal in its current form, we'll await a revised version (and vote on it at the January meeting, if it's received in time).

PR #81, opened 2021-11-14
meeting 22 agenda; updates to attendees.md

SethTisue:


agendas/022-2021-q3.md
```diff
@@ -0,0 +1,57 @@
+# Scala Center Advisory Board Meeting, Q3, 2021: Agenda
+
+## November 8, 2021
```
sjrd, line 3: Traditionally, it seems we have only published the final agenda, and not the possible exchanges that led to it. Is there any reason to do it differently here?

SethTisue: You’re right. I’ll move the extras to invitations .

SethTisue:
@sjrd re-review?

sjrd (approved):

PR #83, opened 2022-01-03
Proposal: Community Delegate Terms

tpolecat:
This proposes that community delegates serve in staggered two-year terms. /cc @bvenners

### DIFF
--- a/proposals/028-community-delegate-terms.md
+++ b/proposals/028-community-delegate-terms.md
@@ -2,7 +2,7 @@
 
 ## Proposer
 
-Proposed by Rob Norris and Bill Venners, community members, 3-Jan 2022.
+Proposed by Rob Norris and Bill Venners, community delegates, 3-Jan 2022.
 
 ## Abstract
 
--- a/proposals/028-community-delegate-terms.md
+++ b/proposals/028-community-delegate-terms.md
@@ -1,3 +1,9 @@
+---
+date: January 2022
+accepted: true
+status: active
+---
+
 # Community Delegate Terms
 
 ## Proposer
--- a/proposals/028-community-delegate-terms.md
+++ b/proposals/028-community-delegate-terms.md
@@ -1,0 +1,29 @@
+# Community Delegate Terms
+
+## Proposer
+
+Proposed by Rob Norris and Bill Venners, community members, 3-Jan 2022.
+
+## Abstract
+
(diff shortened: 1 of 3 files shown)


SethTisue:
attn @ckipp01

who's _probably_ watching this repo ;-)

SethTisue:
Fwiw, this seems good to me. We don't want community representatives to feel like they're on the hook for indefinite service; stepping down after a while should be considered normal. (But also, we might also be happy to keep someone on if they're happy to stay on.)

SethTisue:
Also, I want to observe that modest proposals like this are welcome. Not every proposal needs to be about the Center embarking on some massive long-term effort.

ckipp01:
> attn @ckipp01
> 
> who's _probably_ watching this repo ;-)

Haha, yes I saw. In the future I'll probably just comment on these just as a means of ensuring anyone that creates one knows that I've seen it and will ensure it gets on the agenda for the upcoming meeting.

With that being said, thanks for the proposal @tpolecat. I've actually wondered a bit about this in the past as well, and I'm actually in favor of setting a term length. Firstly, it's pretty common that a standard "length of service" is set for these types of roles. I can quickly think of a few benefits this brings, some that you already hit on

- fresh perspectives
- potentially more incentive to someone to commit knowing there is an end date
- potentially more active members

The last one is a bit vague, but I also think that will long standing members there can be a tendency to just coast, whereas at times when new member join the "freshness" of the role drives them to really be active. On the flip side, there are also benefits to longevity as well.

Again, thanks for bringing this up and I'll make sure it's on the agenda for the upcoming meeting.

ckipp01:
> Also, I want to observe that modest proposals like this are welcome. Not every proposal needs to be about the Center embarking on some massive long-term effort.

Absolutely. There have been times I've though, o, the Center should maybe look into this or put some attention on it, only to deter creating a proposal since it didn't seem "large" enough to warrant one. I'll make a note about this and briefly add a short reminder to reiterate this in the next meeting.

ckipp01:
I went ahead and added a commit on top of this @tpolecat to reflect that it's been accepted during the last meeting. Plus I needed to rebase as there was some changes in the last month to auto-generate the overview page.

All that to say, SCP-028 has been accepted.

ckipp01 (approved):

sjrd (approved):

Issue #85, opened 2022-01-10
The current "status" of proposals


ckipp01:
While looking through the various proposals today I realized that there are quite a few that the "status" in the [proposals readme](https://github.com/scalacenter/advisoryboard/blob/main/proposals/README.md) hasn't been updated in quite some time. I think it'd be a good idea to do a quick review of these and to update the status in the relevant places. I'll provide a table down below of the existing proposals that "remains active" after quite some time. If need be we can discuss this during the next AB meeting, but if we're able to just take care of it this way, then no need.

| Proposal        | Current Status           | Notes  |
| ----------------|:--------------------------:|---------------|
| [002-dotty-migration-path](https://github.com/scalacenter/advisoryboard/blob/main/proposals/002-dotty-migration-path.md) | ~Remains active~ Updated to "completed" | ~There have been tools like [scala3-migrate](https://github.com/scalacenter/scala3-migrate) and a [migration guide](https://docs.scala-lang.org/scala3/guides/migration/compatibility-intro.html). Do we feel this sufficient to mark this as "completed"?~ |
| [004-sip-and-slip-coordiation](https://github.com/scalacenter/advisoryboard/blob/main/proposals/004-sip-and-slip-coordination.md) | Remains active | It says @darja is the SIP coordinator. Is this still active as intended? |
| [005-continuity-of-scala-js](https://github.com/scalacenter/advisoryboard/blob/main/proposals/005-continuity-of-scala-js.md) | ~Remains active~ Updated to "completed" | ~I'm assuming this is still correctly marked as "active".~|
| [008-websites](https://github.com/scalacenter/advisoryboard/blob/main/proposals/008-websites.md) | ~Remains active~ Updated to "completed" | ~I'm actually not 100% sure if this is the case or not?~ |
| [017-lsp-stp-wg-support](https://github.com/scalacenter/advisoryboard/blob/main/proposals/017-lsp-stp-wg-support.md) | ~Remains active~ Updated to "completed" | ~While maybe fluid in its definition, I suppose the continue support of Metals means this is correctly "active"?~ |
| [021-zinc-improvements](https://github.com/scalacenter/advisoryboard/blob/main/proposals/021-zinc-improvements.md) | Remains active | The initial timescales outline in here are far exceeded. So what would we like to do with this? |
[...]

PR #86, opened 2022-01-10
feature: automate the creation of proposals/README.md

ckipp01:
While going over the current status of all the proposals and creating
https://github.com/scalacenter/advisoryboard/issues/85 I figured we
could maintain the proposals/README.md a bit different. I can imagine
it's easy to forget to update that so I think instead it might be a good
idea to instead just use headers in the proposal files themselves. This
way all updates to a proposal and to the current status is just done in
the actual proposal, and then the README can get generated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.scala/
+.bsp/
+.metals/
--- a/README.md
+++ b/README.md
@@ -30,12 +30,35 @@ Proposals should follow the format and sections laid out in the [template
 proposal](https://github.com/scalacenter/advisoryboard/tree/main/templates/proposal.md),
 and should be concise enough to fit on a single side of paper if printed out.
 
-## Recommendations
+### Recommendations
 
 Once a proposal has been adopted by the Advisory Board, it will become a
 recommendation, and should be copied, noting any amendments, into the
 [recommendations](https://github.com/scalacenter/advisoryboard/tree/main/recommendations)
 directory.
+
+### Proposals status
+
+To see the up-to-date status of a proposal see the
+[proposals/README](./proposals/README.md) which holds a summary of updates and
+statuses of all proposals.
+
+_NOTE_: This proposals/README.md file is auto-generated, so if you need to add
+an update or change the status of a proposal make sure to do so in the heading
+of the proposal file. The possible headings are:
+
+```yaml
+date: date proposed
+accepted: true, yes, false, no
+updates:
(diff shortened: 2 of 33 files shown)


ckipp01:
Looks like I don't have permissions to assign reviewers but maybe @SethTisue @sjrd thoughts on this?

SethTisue:
LGTM. (Besides, if we ever decide it was a failed experiment, we can just toss the script and go back to hand-maintenance.)

SethTisue:
coolness points for using scala-cli 😎

sjrd:
I don't have any objection.

@ckipp01 FYI I am currently on sickness leave, so please forgive my unresponsiveness. If something's blocked on me, feel free to ping `@darjutak` and/or `@julienrf`.

julienrf:
Thank you Chris for submitting the PR.

I must say that I am on the fence, though. I am sympathetic to moving the status information of the proposals to the files instead of in the README, however I am a bit worried that we may misspell something in the YAML front matter and that would break the Scala script that generates the README. I am not sure this is a huge win, to be honest, but I am happy to merge it if the others think it should be!

ckipp01:
>  I am a bit worried that we may misspell something in the YAML front matter and that would break the Scala script that generates the README

If this is the primary worry, let me add a check in the script so that if it comes across an unknown key it fails and fails loudly with a message of exactly what's wrong.

I don't have super strong feelings on this, and sort of just did it on a whim 😆. However I do think it's a nice way to manage it. Let me add the check in and then I'll ping you for another review.

ckipp01:
Alright, so I updated it. Now if you do make a mistake and use a key that we don't account for or you accidentally misspell something you'll see:

```
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
"acepted" in file: 001-native-scala-for-spark.md is not an accepted front matter.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

The accepted values are:
  date, accepted, updates, status
```

This should ensure that when you try to generate it and make a mistake it's right in your face. Wdyt @julienrf ?

julienrf:
Nice, feel free to merge the PR!

ckipp01:
> Nice, feel free to merge the PR!

:) I don't have permissions to

PR #87, opened 2022-01-13
Update various proposal statuses

ckipp01:
See the individual commits to each proposal that was updated. This addresses the points in #85 that imo had clear updates on the statuses that we can reflect here.


sjrd (approved):

julienrf (approved):
Thank you Chris!

PR #103, opened 2022-06-28
new technical director

SethTisue:
(in case anyone's wondering: Seb remains at the Center, he's just passing the baton on his administrative role)


ckipp01 (approved):
Ah good catch, thanks!

PR #154, opened 2024-05-14
Propose guidance on choosing Scala between Scala LTS and Next.

zainab-ali:
@eed3si9n , let me know what you think.

### DIFF
--- a/proposals/032-scala-version-guidance.md
+++ b/proposals/032-scala-version-guidance.md
@@ -1,0 +1,41 @@
+---
+date: May 2024
+---
+
+# Provide guidance on choosing between Scala LTS and Next
+
+## Proposer
+
+Proposed by Zainab Ali, Community Representative, 14th May 2024.
+
+## Abstract
+
+There is no clear guidance on how to choose between Scala LTS or Scala Next. This has lead to some open source projects adopting Next where LTS is recommended. Examples include [xebiafunctional/fetch](https://github.com/xebia-functional/fetch/commit/60b26afb242b1e248615c811f2c23a51d6cab6fe), [kovstas/fs2-throttler](https://github.com/kovstas/fs2-throttler/commit/e265f6d585aa55d101ddd5e631abb940fccce084) and [cquiroz/scala-java-locales](https://github.com/cquiroz/scala-java-locales/issues/490).
+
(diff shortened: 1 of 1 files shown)


eed3si9n:
@zainab-ali Great idea.

Potentially something we can get inspiration from are:
- [The Rust Edition Guide](https://doc.rust-lang.org/edition-guide/)

proposals/032-scala-version-guidance.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Proposer
+
+Proposed by Zainab Ali, Community Representative, 14th May 2024.
+
+## Abstract
+
+There is no clear guidance on how to choose between Scala LTS or Scala Next. This has lead to some open source projects adopting Next where LTS is recommended. Examples include [xebiafunctional/fetch](https://github.com/xebia-functional/fetch/commit/60b26afb242b1e248615c811f2c23a51d6cab6fe), [kovstas/fs2-throttler](https://github.com/kovstas/fs2-throttler/commit/e265f6d585aa55d101ddd5e631abb940fccce084) and [cquiroz/scala-java-locales](https://github.com/cquiroz/scala-java-locales/issues/490).
+
+The most prominent guidance in [the 2022 blog post on compatibility plans](https://www.scala-lang.org/blog/2022/08/17/long-term-compatibility-plans.html#the-future-best-practices) is outdated. [The 3.4.0 release notes](https://www.scala-lang.org/blog/2024/02/29/scala-3.4.0-and-3.3.3-released.html#-so-which-version-should-i-update-to) have better guidance, but are hard to discover.
```
eed3si9n, line 15: I think the problem with the 2022 isn't that it's outdated, but it's confusing because it's inconsistent in its messaging.

First it says:

> The compiler team believes that the good answer for this question is always the newest one.

But in the later section it says:

> ## Owners of commercial projects
>
> We suggest you stick to the LTS version.
>
> ## Library maintainers
>
> Unless your library is built around some new language feature available only on Next, you should stick to the latest LTS version to have the broadest possible user base.

So while not everyone are owners of commercial projects or library maintainers, those two are the main active drivers of the Scala community (I think?), so it's confusing.

zainab-ali: This messaging around commercial projects is different to the [release notes](https://www.scala-lang.org/blog/2024/02/29/scala-3.4.0-and-3.3.3-released.html#-so-which-version-should-i-update-to)
> If you are working on a project that is not meant to be used as an external dependency, feel free to update to Scala 3.4.0, especially if you are starting a new project.

This implies that new commercial projects should use `3.4.x` as opposed to LTS.

SethTisue:
merging because the proposal text seems fine

new PR suggesting the proposal be considered completed now: https://github.com/scalacenter/advisoryboard/pull/164

PR #159, opened 2024-08-22
Propose deprecation of Eclipse Scala IDE

zainab-ali:


lrytz:
It's mostly my fault that this didn't happen, I dropped the ball on it. https://users.scala-lang.org/t/scala-ide-for-eclipse/9095.

What I *did* do is organize the domain transfer from Lightbend to EPFL, that is done (cc @fsalvi to confirm).

So I certainly agree to the proposal. I think we can take the website down, not only `noindex` it.

fsalvi:
Yes, the Scala Center own the scala-ide.org
Right now, the website points to the scala-ide github website

SethTisue:
we didn't formally vote at the September meeting, but we did discuss it, and it seemed clear that the proposal should be considered accepted by acclaim

SethTisue:
for the details of what's needed for this to be considered complete, let's move to the new PR: https://github.com/scalacenter/advisoryboard/pull/165

PR #164, opened 2024-10-23
proposal 32 is completed

SethTisue:
we didn't formally vote on SCP-032 at the September meeting, but we did discuss it, and it seemed clear that the proposal should be considered accepted by acclaim

last week, https://scala-lang.org/development/ was published

@zainab-ali do you agree the proposal should therefore be considered completed, or do you think there's still work to be done here?

@sjrd @Gedochao any thoughts?

### DIFF
--- a/proposals/032-scala-version-guidance.md
+++ b/proposals/032-scala-version-guidance.md
@@ -1,5 +1,7 @@
 ---
 date: May 2024
+accepted: true
+status: completed
 ---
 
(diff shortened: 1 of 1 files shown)


sjrd:
Looks good to me.

zainab-ali:
Thanks! We can consider this completed.

sjrd (approved):

